PR #1, opened 2015-06-03
removed X dependency for plot; KGraph parameters.

aaalgo:
Erik,

Thanks for making the benchmark available.  I realized that the installation of KGraph is the most complicated one and I'll make it my next goal to simplify the installation process.

I've made the following changes to your code:
- Switched matplotlib backend to Agg so we no longer need X to plot.
- Specified a universal set of indexing parameters for KGraph.  I've also updated KGraph default parameters to match these.
- Only query parameter P needs to be changed to produce the curve.

Best regards,

- Wei Dong

### DIFF
--- a/ann_benchmarks.py
+++ b/ann_benchmarks.py
@@ -95,15 +95,14 @@ class NearPy(BaseANN):
 
 
 class KGraph(BaseANN):
-    def __init__(self, P, L):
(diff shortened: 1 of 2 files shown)


erikbern:
Thanks! I will re-trigger the benchmarks (currently running on the full GloVe set with 1.15M vectors)

PR #3, opened 2015-06-04
Added NearPy multi-hash configuration and optimized parameters a bit

pixelogik:
I fixed a bug in NearPy cosine similarity, updated it on pypi to 0.2.1. 

I also added multi-hash configuration to the NearPy benchmark, because this is the only way I ever use it, hope that is valid with respect to the benchmark.

### DIFF
--- a/ann_benchmarks.py
+++ b/ann_benchmarks.py
@@ -109,13 +109,19 @@ class PANNS(BaseANN):
 
 
 class NearPy(BaseANN):
-    def __init__(self, n_bits):
+    def __init__(self, n_bits, hash_counts):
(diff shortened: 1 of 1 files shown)


erikbern:
Thanks! Will rerun NearPy!

pixelogik:
Awesome! Thanks! Looking forward to the results. I think NearPy will never be close to the really good libraries, because it is more of a framework for evaluating and comparing ANN methods, than a optimized solution. Still I would like to get it up in the graph a bit, of course.

erikbern:
I think Nearpy might do well on huge data sets (say 1B) though, would be interesting to see

PR #4, opened 2015-06-06
New NearPy configurations, after fixing stuff in NearPy

pixelogik:
I changed two things in NearPy. The really awful thing was, that no UniqueFilter was used by default. That caused the same indexed vector to be multiple times in the search candidate set. After computing the distance to the query vector and sorting with respect to that, in many cases this caused the cropped N=10 nearest neighbour set to contain the same vector multiple times. Also distance computation was more expensive because of the duplicates. The other thing is that I now store normalized vectors in the index which speeds up distance computation a lot. 

Both changes are released via pypi in NearPy 0.2.2

I also ran your benchmark code with a reduced dataset to look for good parameters. 

This example with just 50000 vectors shows that the above fixes helped a lot so far. Especially the missing UniqueFilter created an upper bound on the accuracy before. 

plot_10bit_50000vectors

### DIFF
--- a/ann_benchmarks.py
+++ b/ann_benchmarks.py
@@ -121,7 +121,7 @@ class NearPy(BaseANN):
             nearpy_rbp = nearpy.hashes.RandomBinaryProjections('rbp_%d' % k, self._n_bits)
             hashes.append(nearpy_rbp)
 
-        self._nearpy_engine = nearpy.Engine(X.shape[1], lshashes=hashes, distance=nearpy.distances.CosineDistance())
+        self._nearpy_engine = nearpy.Engine(X.shape[1], lshashes=hashes)
 
         for i, x in enumerate(X):
             self._nearpy_engine.store_vector(x.tolist(), i)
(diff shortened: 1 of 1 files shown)


erikbern:
This is great. I'm off the grid all weekend but will run some more stuff
soon!

I also reached out to the author of LSHForest but haven't heard back -
think there might be some issue with how I'm running it.

pixelogik:
Cool thing! Looking forward to the results. Enjoy the grid-off :)

erikbern:
I've been trying to run this but the EC2 instance gets unresponsive after a few hours...not able to reboot either.

Is it possible that you're using a lot of memory? I'm going to try to add some ulimit stuff and cap it to say 24GB. Should help

pixelogik:
Argh. Let's remove the configurations with the 8 as first parameter then. The lower the number, the more memory is used. Might already help.

pixelogik:
Sorry for the hassle. If that is still a problem I have to refactor a bit to optimize for memory. Know already how.

pixelogik:
Or wait! Let me do a friendlier NearPy conf array. This one is really heavy. Will do another one after lunch quickly that will also look nicer on the graph because more points, steadily increasing in performance.

erikbern:
I added some limits to the code so will try to run again.

https://github.com/erikbern/ann-benchmarks/blob/master/ann_benchmarks.py#L14

pixelogik:
Ok. If that still has problems and you still have time for that, would be cool if you could try this, which is less aggressive on memory and only variing one parameter: 
'nearpy': [NearPy(16, 5), NearPy(16, 10), NearPy(16, 15), NearPy(16, 20), NearPy(16, 25), NearPy(16, 30), NearPy(16, 40), NearPy(16, 50), NearPy(16, 70), NearPy(16, 90), NearPy(16, 120), NearPy(16, 150)]

erikbern:
The new resource constraints should kill bad behaving child processes (hopefully), so I'll just add your new stuff and will let it run for a while. I'm using this as an excuse to learn Docker so it's all good

pixelogik:
Cool, thx!

erikbern:
Hmm EC2 host unresponsive after the memory limits. Not sure what's going on. Could also have something to do with Docker.

pixelogik:
Hm. Try removing all my old configs, just using the new one from above with 16 bit for all items. That one should not be problematic.

erikbern:
Running again with a lower (12GB) mem limit as well as no configurations below 14 bits

erikbern:
pushed some new stats/plots

running some more nearpy variations

erikbern:
FYI it seems like your code is using Euclidean distance or maybe I'm misunderstanding the defaults! That would explain why it maxes out at 0.7 or something like that

pixelogik:
Thanks for the updates. That is at least better than before. 
Hm, the version of NearPy I released for this pull request (0.2.2) is using CosineDistance as default. Weird. Gonna try some benchmarks on my machine later to see what's up with that.

erikbern:
Hm I might be using the wrong version because I'm running Euclidean
benchmarks now and NearPy is performing pretty well (didn't change anything
about the code):
https://github.com/erikbern/ann-benchmarks/blob/master/results/sift.txt

I'm probably going to re-run everything in a few weeks anyway so no big
deal. Some other package authors reached out to me and wanted to make some
changes

2015-06-09 1:55 GMT-04:00 Ole Sprause <notifications@github.com>:

pixelogik:
Hm ok. Weird. Well however, the current results are already much better than at start. I will experiment with the benchmarking code in the next weeks. 

Thanks again for this cool project!

PR #5, opened 2015-06-11
use BLAS in brute force (via NumPy)

piskvorky:
Replace slow brute force algo (sklearn) by a direct, fast BLAS call.

Make sure you have a fast BLAS installed -- a recent ATLAS, or OpenBlas, Intel's MKL, Apple's Accelerate etc. This can make a huge difference in performance.

PR **not tested** at all. @erikbern can you run it on the AWS machine? I just wrote the code, I didn't get a chance to run it (install assumes Debian). Sorry for any typos.

### DIFF
--- a/install/bruteforce.sh
+++ b/install/bruteforce.sh
@@ -1,0 +1,3 @@
+sudo apt-get install -y python-pip python-dev
+sudo apt-get install -y libatlas-dev libatlas3gf-base
+sudo apt-get install -y python-numpy
--- a/ann_benchmarks.py
+++ b/ann_benchmarks.py
@@ -195,13 +195,12 @@ class BruteForce(BaseANN):
 
     def query(self, v, n):
         """Find indices of `n` most similar vectors from the index to query vector `v`."""
+        # HACK we ignore query length as that's a constant not affecting the final ordering
         if self._metric == 'angular':
-            query = v / numpy.sqrt((v ** 2).sum())  # normalize query to unit length
-            cossims = numpy.dot(self.index, query)  # cossim = dot product over normalized vectors
-            dists = -cossims  # just for convenience, so that lowest = best
+            # argmax_a cossim(a, b) = argmax_a dot(a, b) / |a||b| = argmin_a -dot(a, b)
+            dists = -numpy.dot(self.index, v)
         elif self._metric == 'euclidean':
-            # HACK we ignore query length as that's a constant not affecting the final ordering:
-            # argmax_a (a - b)^2 = argmax_a a^2 - 2ab + b^2 = argmax_a a^2 - 2ab
+            # argmin_a (a - b)^2 = argmin_a a^2 - 2ab + b^2 = argmin_a a^2 - 2ab
             dists = self.lengths - 2 * numpy.dot(self.index, v)
         else:
             assert False, "invalid metric"  # shouldn't get past the constructor!
--- a/ann_benchmarks.py
+++ b/ann_benchmarks.py
@@ -22,7 +22,7 @@ if soft == resource.RLIM_INFINITY or soft >= memory_limit:
 class BaseANN(object):
     pass
 
-        
+
 class LSHF(BaseANN):
     def __init__(self, metric, n_estimators=10, n_candidates=50):
         self.name = 'LSHF(n_est=%d, n_cand=%d)' % (n_estimators, n_candidates)
@@ -118,7 +118,7 @@ class PANNS(BaseANN):
         self._n_trees = n_trees
         self._n_candidates = n_candidates
         self._metric = metric
-        self.name = 'PANNS(n_trees=%d, n_cand=%d)' % (n_trees, n_candidates)        
+        self.name = 'PANNS(n_trees=%d, n_cand=%d)' % (n_trees, n_candidates)
 
     def fit(self, X):
         self._panns = panns.PannsIndex(X.shape[1], metric=self._metric)
@@ -175,17 +175,24 @@ class KGraph(BaseANN):
 
 
 class BruteForce(BaseANN):
+    """kNN search that uses a linear scan = brute force."""
     def __init__(self, metric):
+        if metric not in ('angular', ):
+            raise NotImplementedError("BruteForce doesn't support metric %s" % metric)
         self._metric = metric
         self.name = 'BruteForce()'
 
     def fit(self, X):
-        metric = {'angular': 'cosine', 'euclidean': 'l2'}[self._metric]
-        self._nbrs = sklearn.neighbors.NearestNeighbors(algorithm='brute', metric=metric)
-        self._nbrs.fit(X)
+        """Initialize the search index."""
+        # normalize vectors to unit length
+        self.index = X / numpy.sqrt((X ** 2).sum(-1))[..., numpy.newaxis]
 
     def query(self, v, n):
-        return list(self._nbrs.kneighbors(v, return_distance=False, n_neighbors=n)[0])
+        """Find indices of `n` most similar vectors from the index to query vector `v`."""
+        query = v / numpy.sqrt((v ** 2).sum())  # normalize query to unit length
+        cossims = numpy.dot(self.index, query)  # cossim = dot product over normalized vectors
+        indices = numpy.argsort(cossims)[::-1]  # sort by cossim, highest first
+        return indices[:n]  # return top `n` most similar
 
 
 def get_dataset(which='glove', limit=-1):
@@ -246,7 +253,8 @@ def get_queries(args):
             print len(queries), '...'
 
     return queries
-            
+
+
 def get_algos(m):
     return {
         'lshf': [LSHF(m, 5, 10), LSHF(m, 5, 20), LSHF(m, 10, 20), LSHF(m, 10, 50), LSHF(m, 20, 100)],
@@ -305,7 +313,7 @@ if __name__ == '__main__':
     if os.path.exists(results_fn):
         for line in open(results_fn):
             algos_already_ran.add(line.strip().split('\t')[1])
-            
+
     algos = get_algos(args.distance)
     algos_flat = []
 
@@ -313,7 +321,7 @@ if __name__ == '__main__':
         for algo in algos[library]:
             if algo.name not in algos_already_ran:
                 algos_flat.append((library, algo))
-                
+
     random.shuffle(algos_flat)
 
     print 'order:', algos_flat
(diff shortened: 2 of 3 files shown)


install/bruteforce.sh
```diff
@@ -0,0 +1,3 @@
+sudo apt-get install -y python-pip python-dev
+sudo apt-get install -y libatlas-dev libatlas3gf-base
+sudo apt-get install -y python-numpy
```
piskvorky: @erikbern I'm not terribly sure how the debian-packaged NumPy plays with BLAS... can you check that BLAS is being picked up by NumPy (=dot calls are fast)?

erikbern: it seems like the pypi version of numpy gets pulled in from some other package so it might not be needed anyway

erikbern:
LGTM except needs support for Euclidean distance as well.

Btw I wonder if Annoy could leverage BLAS for fast dot products etc... probably means I have to rewrite some of it in matrix form

piskvorky:
Hmm, I could have sworn I read "tests are cossim only" in the README Principles, but I don't see it there now.

Will try to add Euclidean when I have time again, but can't promise any ETA :(

Btw, as a sanity check, what is the time for `numpy.dot(x, x.T)` for `x = numpy.random.rand(1000, 1000)` on the "test machine"?

erikbern:
I changed the benchmarks a bit so that both cosine and Euclidean are represented

piskvorky:
I ran some of the tests on our machines (~dedicated server, not AWS). The PR seems to work, no typos :)



These look a bit different to your results, not sure why. It's possible I messed something up, it's not clear to me what the launch process is (I just deleted all files under `./results` and then ran `ann_benchmarks.py`).

In particular, everything except FLANN, ANNOY and KGraph seems to perform worse than brute force now, on the GloVe dataset...

piskvorky:
Btw the chosen colour palette of the plot is a benchmark on its own (for my eyesight) :)

maciejkula:
What surprises me is that LSHF does so badly. Does the author know about this benchmark? It's such a recent addition to sklearn.

maciejkula:
@maheshakya pinging you in case you'd like to do some parameter tweaking on this benchmark?

erikbern:
Thanks for running!

I think it's possible I'm doing something weird with LSHF but need to
figure out.

On benchmark discrepancy: might be caused by different compiler settings
etc. I might look at it later. In particular Annoy doesn't use any compiler
optimizations so I want to enable some :)

On vacation for another week so this might take a while.

Btw do you think it makes sense to look at bigger data sets?

hectorgon:
Have you tried using hierarchical kmeans / svd (using the eigenvectors as splitting planes)?
When I experimented with video embeddings they seemed to do much better than LSH,
at least for embeddings trained using dot / cosine distance loss functions.
The key thing that got the performance up for us was adding back-tracking to check nearby partitions.

http://www.jmlr.org/proceedings/papers/v28/weston13.pdf

piskvorky:
OK, I've added the Euclidean metric as well.



(I only ran on Annoy and brute force because they're fast to build -- there were no changes to other algos = their relative performance stays the same)

I also noticed that the bottleneck is now NOT the distance computations, but rather sorting for "k nearest nerighbours" at the very end :) So I optimized that too (assumes NumPy >= 1.8):



The brute force algo is now practically on par with its implementation in gensim. So it should be a worthy baseline for all the ANNs :)

install/bruteforce.sh
piskvorky: Alright... can you check the timings for `np.dot` anyway, just to be sure? What version of numpy is that?

searchivarius:
@piskvorky does BLAS create its own copy of the data?

piskvorky:
@searchivarius I know `numpy.dot` sometimes did memory copies, when the matrix order was "wrong" (Fortran order vs C order = column major vs. row major). For this reason, I used to call GEMM manually in HPC apps (via `scipy.fblas`), setting the GEMM transposition flags by hand, rather than risk any silent memory copies.

I actually [asked a question about this, many years ago](http://stackoverflow.com/questions/6061720/array-order-in-numpy-dot), but there was no good answer.

In any case, this is mostly relevant for level 3 BLAS (GEMM = matrix * matrix multiplications etc). This benchmark uses only level 2 (matrix * vector), and I'm pretty sure NumPy always does the sane thing, no matter the underlying BLAS implementation.

piskvorky:
One more optimization for brute force: switched to single precision math. About 50% speed up:

searchivarius:
I haven't even bothered about doubles. Is there any chance that BLAS uses multiple threads? And what is the server that you are using?
Thanks!

piskvorky:
@searchivarius [server](https://www.hetzner.de/lt/hosting/produkte_rootserver/ex40ssd). And yes, most BLAS implementations use threads internally.

searchivarius:
@piskvorky this makes a lot of sense. If you have 8-16 cores, this can easily compensate for the random memory layout. However, this doesn't make a fair comparison.

piskvorky:
@searchivarius The machine has 4 cores. And all algorithms run on the same machine -- isn't that the point of the benchmark?

searchivarius:
I don't think so:
```
Use single core benchmarks. I believe most real world scenarios could be parallelized in other ways (eg. do multiple queries in parallel).
```

searchivarius:
Well, some algorithms might not scale well with the number of threads (though not very likely). However, it is always possible to test this by running a multi-threaded benchmark, which, however, executes a single-threaded implementation.

piskvorky:
Oops, good catch @searchivarius ! Must be one of the newer Principles I missed, I don't remember reading that earlier.

Anyway, dumbing down BLAS would be too hard to do in general, I'm not going there. I'll leave this PR as is -- up to Erik. It's possible the speed-up is only 10x then :-)

maciejkula:
Does annoy use floats or doubles for this benchmark?

erikbern:
I would prefer it if we can dumb down BLAS so it uses only a single core.

Alternatively we can change the benchmark so it uses whatever many cores are available on the machine.

aaalgo:
The speedup of BLAS mainly comes from lowering cache hit rate with blocking.  If the query matrix is reasonably large (batch size * dimension), the speedup achieved by batch processing could be more than that achieved by multi-processing.  Such batch processing is typically not exploited in other k-NN algorithms.  That somehow gives an edge to brutal force with BLAS even without parallelization, as not all k-NN applications allow batch processing.  I think this is OK since everyone is trying to beat brutal force by a few times, but it will definitely help to keep the original version of brutal force.

The performance gap between different BLAS implementations could be huge.  As I last benchmarked it, OpenBLAS is better than ATLAS, and is reasonably close to MKL (http://www.wdong.org/wordpress/blog/2013/08/30/mkl-vs-atlas-vs-openblas/).
I would stick to OpenBLAS for benchmarking purpose.
On Ubuntu 14.04, when libopenblas-dev is installed, numpy links to openblas by default.  This post explains how to manually tune level of parallelization http://stackoverflow.com/questions/29559338/set-max-number-of-threads-at-runtime-on-numpy-openblas .  Note that if it works, controlling number of threads with the environment variable OMP_NUM_THREADS should be the most straightforward approach (for example,  with OMP_NUM_THREADS=4 python xxx.py), but one could manually build OpenBlas without OpenMP.

piskvorky:
@aaalgo agreed on all points. Btw gensim does do batch processing of brute force (BLAS) transparently, when querying by multiple documents.

Someone can update this PR to use OpenBlas instead of ATLAS, and limit the #threads to 1. Should be a simple ~3 line change, but I won't have time to play with this, so not me.

aaalgo:
I can do what I suggested if this pull request is merged into the main branch.

erikbern:
I can merge if you rebase please

aaalgo:
Just want to clarify that the pull request is not from me.

erikbern:
yes I was hoping @piskvorky can rebase :)

i can do it if necessary though

piskvorky:
This seems to have died; I still think fast pure-Python similarities (BLAS via NumPy) are a worthy baseline: simple, easy to deploy, maintain, fast.

I won't have time for this, but if @aaalgo wants to add the OpenBlas threading restrictions, that would be cool (I didn't realize the contenders must be single threaded, sorry).

erikbern:
I'm also happy to change all benchmarks to be multi-threaded (in that case
I'll just use a thread pool to do multiple nearest neighbor searches
simultaneously for all libraries)

2015-08-26 20:53 GMT-04:00 Radim Řehůřek <notifications@github.com>:

piskvorky:
I think that makes good sense.

I love that these ANN benchmarks are practical -- practical datasets, practical lib installs, practical implementations. A practical, reproducible HW setup fits the theme nicely IMO.

aaalgo:
I don't write the fastest Python code, but I'll add a parallel BLAS mode in my KGraph API just for the purpose of benchmark.  Multi-threading can be enabled in KGraph by passing in the "threads=8" parameter in the python search API.  It should be a little bit faster than an external thread pool.

aaalgo:
I have updated my KGraph repository.  After rebuilding the source, BLAS mode can be enabled by passing "blas=True" as in "index.search(dataset, query, K=K, blas=True)".  No need to call index.build if only brute force with BLAS is to be used.  Speedup will only start to show as dimension is > 100.

erikbern:
i merged this as a separate algo that is used to compute the correct results

 seems to be like 100x faster than before :)

searchivarius:
It could be, b/c Python brute-force is 10x slower than a single-thread brute-force.

piskvorky:
Hooray! \o/ Thanks @erikbern.

Feel free to add a note that people can bug me regarding this code -- I'll be happy to maintain it, in case of any bugs / questions / extensions.

piskvorky:
@erikbern have the graphs on the main README page been regenerated?

The numbers seem dodgy (brute force slower than FLANN on 100% acc). What BLAS was this using?

erikbern:
brute force doesn't use BLAS in the benchmarks

piskvorky:
Do you mean you're letting NumPy link against whatever BLAS is already installed in your system, or you're specifically disabling BLAS during NumPy installation?

erikbern:
https://github.com/erikbern/ann-benchmarks/blob/master/ann_benchmarks/__init__.py#L331

see `BruteForce` vs `BruteForceBLAS`

erikbern:
the reason I don't want to use BruteforceBLAS for benchmarks is that it uses multiple CPU cores by default and I'm not sure how to disable that.

piskvorky:
Aah, sorry, I thought BruteForce == BruteForceBLAS, for some reason. Never mind :)

erikbern:
i might rewrite the benchmarks so it runs on multiple threads instead... that way i don't have to worry about it. more realistic too

PR #10, opened 2015-11-30
adjust glove.sh to the new URLs for the GloVe datasets

ilyaraz:
GloVe datasets got moved, this is a fix to download them from the right place.


erikbern:
Thanks!

I actually uploaded individual files to `https://s3-us-west-1.amazonaws.com/annoy-vectors/glove.twitter.27B.100.txt.gz` etc. so it's even better if we can use them. but will merge this for now and I can update later today

erikbern:
KGraph isn't building but I can fix separately

PR #15, opened 2016-03-14
getting a dataset in a separate subprocess

ilyaraz:
Since parsing glove.txt takes lots of memory, it should be done in a separate subprocess so that this memory is getting consistently released.


erikbern:
separately we should support just reading files incrementally and not loading the entire data set into memory.

PR #16, opened 2016-03-17
Adding FALCONN

ilyaraz:
I've added FALCONN (http://falconn-lib.org/) tuned for the GloVe dataset.

Here are the results for GloVe on the dedicated c4.2xlarge instance: https://gitlab.com/ilyaraz/experiments/blob/master/glove_results/ec2_flat.txt (other numbers are taken from the upstream of ann-benchmarks).

Here is the comparison of Annoy vs. FALCONN in one point *on the same instance* (to make sure that the numbers from the above make sense): http://pastebin.com/gpDVFQhm

### DIFF
--- a/ann_benchmarks/__init__.py
+++ b/ann_benchmarks/__init__.py
@@ -49,7 +49,7 @@ class FALCONN(BaseANN):
         self._params.storage_hash_table = 'flat_hash_table'
         self._params.seed = 95225714
         self._index = falconn.LSHIndex(self._params)
-        self._index.fit(X)
+        self._index.setup(X)
         self._index.set_num_probes(self._num_probes)
         self._buf = numpy.zeros((X.shape[1],), dtype=numpy.float32)
 
--- a/ann_benchmarks/__init__.py
+++ b/ann_benchmarks/__init__.py
@@ -1,4 +1,4 @@
-import gzip, numpy, time, os, multiprocessing, argparse, pickle, resource, random
+import gzip, numpy, time, os, multiprocessing, argparse, pickle, resource, random, math
 try:
     from urllib import urlretrieve
 except ImportError:
@@ -18,6 +18,48 @@ class BaseANN(object):
     pass
 
         
+class FALCONN(BaseANN):
+    def __init__(self, metric, num_bits, num_tables, num_probes):
+        self.name = 'FALCONN(K={}, L={}, T={})'.format(num_bits, num_tables, num_probes)
+        self._metric = metric
+        self._num_bits = num_bits
+        self._num_tables = num_tables
+        self._num_probes = num_probes
+        self._center = None
+        self._params = None
+        self._index = None
+        self._buf = None
+
+    def fit(self, X):
+        if X.dtype != numpy.float32:
+            X = X.astype(numpy.float32)
+        if self._metric == 'angular':
+            X /= numpy.linalg.norm(X, axis=1).reshape(-1,  1)
+        self._center = numpy.mean(X, axis=0)
+        X -= self._center
+        import falconn
+        self._params = falconn.LSHConstructionParameters()
+        self._params.dimension = X.shape[1]
+        self._params.distance_function = 'euclidean_squared'
+        self._params.lsh_family = 'cross_polytope'
+        falconn.compute_number_of_hash_functions(self._num_bits, self._params)
+        self._params.l = self._num_tables
+        self._params.num_rotations = 1
+        self._params.num_setup_threads = 0
+        self._params.storage_hash_table = 'flat_hash_table'
+        self._params.seed = 95225714
+        self._index = falconn.LSHIndex(self._params)
+        self._index.fit(X)
+        self._index.set_num_probes(self._num_probes)
+        self._buf = numpy.zeros((X.shape[1],), dtype=numpy.float32)
+
+    def query(self, v, n):
+        numpy.copyto(self._buf, v)
+        if self._metric == 'angular':
+            self._buf /= numpy.linalg.norm(self._buf)
+        self._buf -= self._center
+        return self._index.find_k_nearest_neighbors(self._buf, n)
+
 class LSHF(BaseANN):
     def __init__(self, metric, n_estimators=10, n_candidates=50):
         self.name = 'LSHF(n_est=%d, n_cand=%d)' % (n_estimators, n_candidates)
@@ -382,6 +424,14 @@ def get_algos(m):
     if m == 'angular':
         # RPForest only works for cosine
         algos['rpforest'] = [RPForest(leaf_size, n_trees) for n_trees in [3, 5, 10, 20, 40, 100, 200, 400] for leaf_size in [3, 5, 10, 20, 40, 100, 200, 400]]
+        L = []
+        x = 1
+        while True:
+            L.append(x)
+            if x >= 1400:
+                break
+            x = int(math.ceil(x * 1.1))
+        algos['falconn'] = [FALCONN(m, 16, l, l) for l in L]
 
     return algos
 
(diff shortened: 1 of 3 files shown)


ilyaraz:
I should probably trick Travis into running our unit tests at some point...

erikbern:
impressive!

ilyaraz:
Thanks!

What's the further protocol? Should someone re-run all the algorithms on a single instance? How many hours does it take, btw?

erikbern:
Sure, feel free to re-run it. It takes like 4 days, IIRC. I wonder if the bug with k-graph has been fixed.

Separately it's time to start tweaking Annoy :)

ilyaraz:
Uh, isn't it like $200 or so? :)

One big optimization would be not to parse the dataset file for every set of parameters. For FALCONN, it would reduce the overall running time by a factor of more than 2, I think.

searchivarius:
I think it makes absolutely no sense to re-run anything slower than FLANN. I would actually not re-run FLANN either as long as we use the same instance. There is some instance-to-instance variability, but really doubt that it is more than 10-30% difference. Some further savings can be obtained by saving indices to disk and reloading them when appropriate.

What would actually be also nice to re-run at the same point in the near future when we all roll out the latest versions. @erikbern do you have more concrete dates relating your tweaked version?

Also, @ilyaraz do you plan to test on SIFT?

Thanks!

ilyaraz:
Yes, we plan to run FALCONN on SIFT at some point. In fact, we did it for the paper, but we updated the code quite a bit since then.

erikbern:
Yeah there's no point re-running the slow ones – you are right. There's some minor improvements to Annoy so I want to re-run that at some point. 

Also it's not that expensive. If you get a spot instance they are usually only a few cents an hour

I can rerun in a month or two maybe?

yurymalkov:
Test results with FALCONN on a different instance http://s30.postimg.org/r9hiov36p/glove.png . Annoy is not the very last version though.

searchivarius:
@erikbern two months will be great. Regarding spot instances, I know about them, but I never used so far, because I am afraid that they will terminate my long-working pipelines and I will have to restart them from scratch. :-)

ilyaraz:
@yurymalkov what is the machine you were using? Also, we you running FALCONN with the same parameters that are currently hardcoded in the `__init__.py` of ann-benchmarks?

yurymalkov:
@ilyaraz I ran it on a Xeon E5-4650 v2 CPU system with 120 Gb of RAM. I have used the version of __init__.py commited by you a few days ago.

erikbern:
spot instances are usually fine – i make sure to copy a snapshot of the results once a day or so. they are a 1/3 of the cost. the weird thing is i think they use a FIFO to kill them or something – every once in a while i end up having a "magic" spot instance that stays online for months

searchivarius:
@yurymalkov I do reproduce Ilya's numbers of my laptop (it's actually only 7% slower than an Amazon instance). Make sure you use all the latest FALCON code and (Ilya's version of the benchmark).

erikbern:
btw seems like FALCON caps out at 98% roughly?

ilyaraz:
The last point for FALCONN is exactly 0.99.

yurymalkov:
@searchivarius I have rerun the FALCONN tests with the latest versions of both the script and the library. Thankfully it is very fast to build) However the results are almost indistinguishable, for example:
**was** 
falconn	FALCONN(K=16, L=408, T=408)	75,21686	0,00386	0,8727
falconn	FALCONN(K=16, L=494, T=494)	85,6264	0,00464	0,9025
falconn	FALCONN(K=16, L=725, T=725)	125,86021	0,00663	0,9524
**now**
falconn	FALCONN(K=16, L=408, T=408)	49,72762	0,0039	0,8727
falconn	FALCONN(K=16, L=494, T=494)	58,05403	0,0044	0,9025
falconn	FALCONN(K=16, L=725, T=725)	83,34483	0,00633	0,9524
There is a difference in construction time due to the number of free threads on the server (which was now fully devoted to testing).

ilyaraz:
Well, I guess this run says that, unsurprisingly, CPU frequency is very important for a single-thread experiments.

erikbern:
Btw does FALCONN use threads during queries? My intent is to limit it to 1 CPU – I tried to configure for that, but never actually verified that it works

ilyaraz:
It uses many threads during the preprocessing, but only a single one during the query phase.

erikbern:
Ok, good. Cheating otherwise :)

searchivarius:
@ilyaraz I am running the tests on a 2-year old laptop. Can't have a great CPU frequency. Yet, my results 1.5x better than @yurymalkov 's and only 7% worse than yours. Doesn't make sense. Maybe, compiler doesn't pick up all of the SIMD available. Or, who knows, may be there is some memory ordering/layout issue.

searchivarius:
@erikbern having 18 cores sitting and use only one core isn't such a good approach :-)

erikbern:
I just don't think it makes sense to use threads for a single query – better to parallelize queries instead. This is easy to do in the benchmark script btw – just create a threading.Pool and run all the queries using that. Since queries are immutable it makes more sense to leave the threading to the user rather than the library

erikbern:
re-running the benchmarks and i just pushed some fresh results that include FALCONN. the results are very impressive for high precisions

not done yet, will take another couple of days

yurymalkov:
@erikbern  Erik, Leonid and I are going to add new algorithms/settings for the NMSlib shortly. This might require another rerun. Can it be done incrementally?

erikbern:
sure, i'll just wipe out the current nmslib results and rerun. let me know when you are done

yurymalkov:
@erikbern Ok, thanks!

searchivarius:
@erikbern do you resample queries every time you run-tests?

erikbern:
no it uses a fixed random seed

https://github.com/erikbern/ann-benchmarks/blob/master/ann_benchmarks/__init__.py#L298

searchivarius:
This is potentially a loophole for overfitting to the data set (and there are only 1000 queries). :-)

erikbern:
True. I can make it 10k instead.

erikbern:
Doubt it matters in practice

searchivarius:
Well, for my piece of software, I increased the number of queries from 500 to 2K and I saw an almost two-fold slowdown.

yurymalkov:
Making 10k queries might also help to avoid the problem of query results potentially staying cached from the previous series of searches. Now this seem to slightly overestimate the performance for some algorithms at low recall. Not a big deal though.
I.e. output for some algorithm for 3 runs:
((1192514, 100), (1000, 100))
('Built index in', 7.398407936096191)
(**0.00015568709373474122**, 0.167)
(**0.0001187889575958252**, 0.167)
(**0.00011595606803894043**, 0.167)

searchivarius:
@yurymalkov L1/L3 cached I guessed, it should stay in memory anyways. Otherwise, it will not be a fair comparison of annoy which, as far as I remember mmaps data. This was probably the reason why we need three tests (warming up mmaped memory).

erikbern:
annoy doesn't use mmap for ann-benchmarks actually. i haven't done any performance testing but i don't think it should affect performance. mmapped data gets cached in L1/L3 too right?

the page cache will shift out any unused pages to the disk after a while so doing it 3 times should help all algorithms

i can increase it to 10k, sure. building indexes is what takes like 95% of the runtime anyway right now

searchivarius:
@erikbern 10K sounds good.

searchivarius:
PS: regarding mmap, I am not sure that after saving to disk and mmaping back saved data, it will all be cached well. I am not even sure it will be in memory. I guess it depends a lot on how smart OS is.

erikbern:
I'm pretty sure mmapped data behaves exactly like data that is allocated. If you allocate a block of 100MB using malloc and you don't use it for a while, it will be stored on disk instead

searchivarius:
I seem to fail to explain this properly. Anyways, this doesn't matter if you don't use mmap.

PR #19, opened 2016-05-21
Using NMSLIB code from a release rather than from an obscure branch

searchivarius:
I have retested NMSLIB using the updated version (which downloads v1.5). Seems to be fine. If you were waiting for an updated version to run SIFT benchmarks, you can run them now. Even faster with saved indices. We don't cheat and don't save answers there :-) Only the indices themselves.

Additionally, I changed the source location of convert_texmex_fvec.py (to be from NMSLIB v1.5 too). I sometimes rename folders, it's not a good idea to pull it from master. Really, on a second thought: ann_benchmark could probably keep its own copy of convert_texmex_fvec.py.

### DIFF
--- a/install/nmslib.sh
+++ b/install/nmslib.sh
@@ -2,8 +2,10 @@ cd "$(dirname "$0")"
 echo "Installing Python interface for the Non-Metric Space Library"
 # Remove the previous version if existed
 rm -rf nmslib 
-# Note that we use the pserv branch here:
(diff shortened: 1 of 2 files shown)


erikbern:
is there a major impact? i can rerun benchmarks

searchivarius:
no impact (I checked). this is merely to ensure that the right version is used.

PR #21, opened 2016-06-09
fixed kgraph, caching train/test data

aaalgo:
- Fixed KGraph performance issue (substantial changes to pykgraph implementation to remove the per-batch query overhead, which cannot be amortized with evaluated with batch-size 1).
- Allow saving and reusing KGraph index. 
- Caching train test data to speedup evaluation.

### DIFF
--- a/ann_benchmarks/__init__.py
+++ b/ann_benchmarks/__init__.py
@@ -213,18 +213,28 @@ class KGraph(BaseANN):
         self._metric = metric
 
     def fit(self, X):
+        os.environ['OMP_THREAD_LIMIT'] = '40'
         import pykgraph
 
-        if self._metric == 'angular':
-            X = sklearn.preprocessing.normalize(X, axis=1, norm='l2')
-        self._kgraph = pykgraph.KGraph()
-        self._kgraph.build(X, iterations=30, L=100, delta=0.002, recall=0.99, K=25)
-        self._X = X # ???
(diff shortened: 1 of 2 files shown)


ann_benchmarks/__init__.py
```diff
@@ -213,18 +213,28 @@ def __init__(self, metric, P):
         self._metric = metric
 
     def fit(self, X):
+        os.environ['OMP_THREAD_LIMIT'] = '40'
```
erikbern: not sure if i love this, but i guess it's fine for now

i'm trying to make up my mind whether we should limit ann-benchmarks to 1 core or not. one idea i've been thinking about is let's not impose any limitations but run `query` using a thread pool – only downside is if thrashing kills the L2/L3 cache

erikbern:
nice!

ann_benchmarks/__init__.py
aaalgo: This is only for index construction which follows existing code of NmslibReuseIndex.  Search is still single threaded  (I'll post about construction time separately.)  
Regarding search time parallelization, KGraph batch search achieves 3.4x speedup on a 4 core CPU with hyperthreading on the glove dataset.

erikbern: Yeah i know it's only for index construction. I think it's fine to use multiple threads for index construction. Maybe I'll add support for that in Annoy too

Issue #23, opened 2016-06-13
Nearpy distance is Cosine by default - Euclidean not applied


erramuzpe:
Hi, 

I was checking some libraries and realized that in the case of Nearpy, the euclidean distance is not applied in the benchmark. As can be seen in the [docs](https://github.com/pixelogik/NearPy#engine) of Nearpy, the default is the cosine but the euclidean one is not applied. I will PR changes
thanks for the bench, it's great

PR #24, opened 2016-06-13
Solving #23

erramuzpe:
Hi Erik,

here there are 2 small changes to solve the #23 , hope it fits you. I have also added the normalization of the data for the angular distance, as it is in other libraries, but not sure if this applies for Nearpy. 

cheers


erikbern:
thanks!

PR #26, opened 2016-06-17
Updated kgraph parameters and etc

aaalgo:
- A command line argument "--algo" is added so one can run only one algorithm instead of all.
- Updated kgraph configuration for two dataset separately.

sift
glove

### DIFF
--- a/ann_benchmarks/__init__.py
+++ b/ann_benchmarks/__init__.py
@@ -214,10 +214,11 @@ class NearPy(BaseANN):
 
 
 class KGraph(BaseANN):
-    def __init__(self, metric, P):
-        self.name = 'KGraph(P=%d)' % P
+    def __init__(self, metric, P, index_params):
+        self.name = 'KGraph(%s,P=%d)' % (metric, P)
         self._P = P
         self._metric = metric
+        self._index_params = index_params
 
     def fit(self, X):
         os.environ['OMP_THREAD_LIMIT'] = '40'
@@ -232,7 +233,7 @@ class KGraph(BaseANN):
         if os.path.exists(path):
             self._kgraph.load(path)
         else:
-            self._kgraph.build(iterations=30, L=100, delta=0.002, recall=0.99, K=25)
+            self._kgraph.build(**self._index_params) #iterations=30, L=100, delta=0.002, recall=0.99, K=25)
             if not os.path.exists(INDEX_DIR):
               os.makedirs(INDEX_DIR)
             self._kgraph.save(path)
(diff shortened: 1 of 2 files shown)


erikbern:
nice! i will re-run all algorithms shortly!

searchivarius:
@erikbern would you wait a few days? :-) we will make some minor updates.

erikbern:
yeah just lmk when you are done. maybe the upcoming weekend

searchivarius:
Ok, thanks! I will also update the code so that it is possible to run with and without saving indices. Saving will be set by default b/c I can see no serious reason to disable it, but you will be able to do it if you wish.

erikbern:
ok sounds good

erikbern:
i'm rerunning the benchmarks but kgraph keeps crashing

the last thing it outputs is `Graph completion with reverse edges...`

searchivarius:
@erikbern I am doing the same now (before submitting a pull request), and I see the same thing. It works on sift (greatly) but crashes on glove.

erikbern:
ok weird. btw i've changed the setup a bit and the results are slightly different. see https://github.com/erikbern/ann-benchmarks/pull/28

searchivarius:
@erikbern I thought Python had no real threads (at least not 2.x). But, can FLANN use threads internally, inside the library?

searchivarius:
PS: by internally, I mean intra-query parallelism.

erikbern:
Python has real threads but also the global interpreter lock. But if you use threads inside a library, there's no locking

searchivarius:
@erikbern that's right, you can run a multi-threaded test in Python only if a library somehow supports multi-threading inside it. For example, it can have intra-query parallelism or it has its own thread pool. To make use of the thread pool, there should be some support for batch-query submission. I don't think most libraries support this. I also think that FALCONN isn't thread safe (though it uses multi-threading for index creation).

searchivarius:
Intra-query || is possible, but it's much easier to have separate threads each processing the complete request (unless you need to reduce latency).

aaalgo:
I'm looking into the problem right now.  Will report shortly.

aaalgo:
I ran out of memory.  I've updated KGraph.   Even though memory is still tight, the benchmark is working now on my c4.2xlarge instance.

searchivarius:
Nice, thanks. I am going to submit a pull request within several hours. The performance is nearly the same, but it has a flag to enable/disable saving of indices + a bit of refactoring to reduce the clutter.

PR #28, opened 2016-06-28
Thread pool

erikbern:
Instead of setting a thread limit, just saturate all CPUs by using a thread pool


erikbern:
results so far: 

image

erikbern:
seems like FLANN is much better, not sure why.

FALCONN is a bit worse, which I think is because it uses Eigen, which runs multithreaded otherwise.

yurymalkov:
@erikbern AFAIK FALCONN search is not thread-safe: https://github.com/FALCONN-LIB/FALCONN/issues/6 .
That might be the reason for low recall.

erikbern:
Ok – that explains it.

Not sure how to deal with it. I feel pretty strongly that the real metric is how much quality you get per CPU cycles. Not using threads is problematic because then a library that's multithreaded behind the scenes can use more CPU cycles than it seems like. Using threads during benchmarks means you can let libraries take advantage of threading and you can compare apples to apples

searchivarius:
@erikbern For these kind of indices, a multi-threaded test will have roughly the same results. At least, this is my experience. You have static indices that fit into memory, so there is no real contention. When you use K threads to run K queries in parallels, the runtime of each query does increase a bit, but not too much.

searchivarius:
Anyways, as @yurymalkov remind us, actually there seems to be a way to release GIL inside the Python extension. I didn't try to do this b/c I couldn't find a proper documentation for this. Perhaps, it's worth investigating how to do this properly. I checked FLANN sources, but I don't understand how it creates bindings. It seems to use some 3rd party package that does all the heavy lifting.

erikbern:
Updated stats:

Looks like BallTree isn't doing so well

image

searchivarius:
The ball tree is a good benchmark only for smaller intrinsic dimensionality. It was also a good benchmark in non-metric spaces where not many benchmarks are available. Now, however, SW-graph/HNSW do better in some of these hard cases (but actually not in all of them!!!).

Anyways, I think it's worth testing only methods that performed very well in a single-thread benchmark. In addition, Annoy and NMSLIB should be modified to release GIL (I hope this is possible without crashing etc). Otherwise, this is not a fair comparison.

erikbern:
I don't think GIL is causing any contention – it doesn't apply when you are calling an external library.

erikbern:
Latest results:

image

Not sure what's going on with FALCONN, maybe it's the lack of thread safety as you mentioned @searchivarius . Would really like to switch to multithreaded benchmarks but not if it breaks libraries

aaalgo:
@erikbern I just did an experiment with KGraph and can confirm that GIL must be explicitly released by extension to allow parallelization with ThreadPool.  Basically the parallel code must be wrapped with the macro pair "Py_BEGIN_ALLOW_THREADS" and "Py_END_ALLOW_THREADS" [Doc](https://docs.python.org/2/c-api/init.html).  The KGraph performance in your current benchmark is still single threaded, and I guess this is the case for some of other libraries.  At least I didn't find Py_BEGIN_ALLOW_THREADS by grepping the code base of annoy and rpforest.

I've added the GIL code to the KGraph repo and there should be about 4x speedup.  I think we need to be sure that either a contender actually supports multi-threading, or it is removed from the parallel benchmark.

yurymalkov:
@erikbern @searchivarius and I agree with @aaalgo. Our tests for the nmslib also indicate that releasing the GIL leads to significant increase of performance.

searchivarius:
Ok, I have incorporated @yurymalkov changes. I wanted to test them, but I noticed a merge conflict in @erikbern code:
```
<<<<<<< 9467c0df281b03597098ef968f180de8bf4d4671
            if self._save_index:
              self._kgraph.save(path)
        os.environ['OMP_THREAD_LIMIT'] = '1'
=======
            self._kgraph.save(path)
>>>>>>> no thread limits, instead use thread pool for queries
```
PS: I will also need to create a pull request to make ann-benchmark pull a new release.

aaalgo:
I think it should be merged this way
```
            if self._save_index:
              self._kgraph.save(path)
```
This branch doesn't set the environment variable OMP_NUM_THREADS.

searchivarius:
@aaalgo thanks, I will then include this change into the pull request.

searchivarius:
Ok, basic testing is done and the pull reuquest is created.

erikbern:
Interesting. I will update Annoy and try again!

erikbern:
Will re-run benchmarks for annoy, kgraph and nmslib.

Thanks for quick updates!

piskvorky:
Cool! How does this compare to BruteForceBlas now, on some good BLAS (MKL)?

erikbern:
Added it to the benchmarks so that we can see :)

piskvorky:
@erikbern what BLAS was this?

yurymalkov:
@piskvorky Judging from the code it seems it was numpy blas. 
IMHO best MKL implementation can get you something like 2x performance of nmslib bruteforce.

piskvorky:
Yes, seeing something standard like MKL or OpenBlas would be interesting.

erikbern:
feel free to send a PR! not going to spend time on it myself – i think it's interesting to compare to a brute force solution, but judging by the results we already see that ANN is several orders of magnitude faster. I would have been concerned if it wasn't the case :)

erikbern:
err ok maybe it's just one order of magnitude faster. but still acceptable

searchivarius:
One order: 99%
Two orders: 90%
Three orders: 80% (common setting for image retrieval I guess)

searchivarius:
But we shouldn't judge from only a single data set.

piskvorky:
I already did, a long time ago :) It's somewhere in here, I guess, unmerged.

erikbern:
I actually took your code and merged it. See BruteforceBLAS. Just not sure what version of BLAS it runs by default!

ludwigschmidt:
Regarding fast brute force searches via optimized BLAS: another relevant thing to try would be CUDA BLAS on a recent GPU. It won't be good for latency but might be quite competitive for batch nearest neighbor in the high accuracy regime.

If I remember correctly, @ilyaraz looked into this a while ago.

piskvorky:
@erikbern I believe the BLAS installation part was not merged, only my Python code: https://github.com/erikbern/ann-benchmarks/pull/5/files#diff-8b5a3ab5c2fbb09267b1a1cc87674480R1

Though I see my PR there used ATLAS; OpenBlas is probably a better choice by now (also a Debian package).

@ludwigschmidt interesting idea :) I don't think the current benchmarks support "batch kNNs" though, it's always one-query-at-a-time.

erikbern:
Ok I can run it with atlas installed at some point!

aaalgo:
Searching individual query is a vector-matrix operation (level-2) which doesn't really benefit much from BLAS, as there's little cache locality to be exploited as compared to matrix-matrix multiplication (GEMM).  If you look into benchmark results like this https://software.intel.com/en-us/intel-mkl/benchmarks#close , level-2 BLAS routines are not even evaluated.
I'll consider nmslib's brutalforce0 a very close approximation of the BLAS results.  I do expect a huge (>10x) speed boost if queries are submitted in batches of 100s, but I think that's too much an assumption to put on applications.

I think evaluation of BLAS and GPU BLAS in batch mode will divert the focus of this project and therefore better go to a separate project.

piskvorky:
@aaalgo BLAS has a huge impact, even on level 2.

Also, while larger batches help, it's not really >10x (it was about 2x in my benchmarks IIRC, for larger query batches).

aaalgo:
You are right: >10x is unrealistic considering nmslib's brutalforce is already hand-vectorized.  
Level-3 MFLOPS is 2-5x over level-2 MFLOPS according to these pages: [Level 2](http://www.mathematik.uni-ulm.de/~lehn/bench_FLENS/BLAS3/GEMV.html), [Level 3] (http://www.mathematik.uni-ulm.de/~lehn/bench_FLENS/BLAS3/GEMM.html).

PR #31, opened 2016-07-05
Slight performance boost and other changes

searchivarius:
The pull request includes the following:
1. A parameter to disable saving indices for kgraph & nmslib (which is true by default).
2. On a related note, I commented out the code that deletes previously created indices. 
3. Some performance improvements for hnsw only (so it doesn't make sense to rerun any other  NMSLIB methods).
4. Now all NMSLIB methods are initialized in a neat loop rather than through a clunky list.


erikbern:
i have been running benchmarks for a few days now – will have to restart :)

but lgtm

searchivarius:
Sorry about the need to restart. But... please, do save indices. It will finish within hours with this feature.

erikbern:
I can't run with the index saving. The machine runs out of disk space immediately. Sorry will turn it off.

aaalgo:
@erikbern I think the default AWS image is designed to only provide the OS, rather than for one to do anything serious with its storage.  It's almost for sure that you'll save a lot more in time cost than the extra storage cost you put in.  Also at some point, say if you intend to add an additional dataset,  you'll be forced to add some additional storage.

erikbern:
yeah probably has to attach an EBS volume, but I don't really have time to look into it right now

searchivarius:
@erikbern I am not in a hurry. Regarding the disk size: I never used Docker, so I just start the spot instance manually, copy data (and sometimes pickled queries), and start the benchmark.

PR #37, opened 2017-10-17
Major refactoring & added features

maumueller:
Hi Erik,

as discussed our changes in one very big pull request. A concise description of our work: 

* Moved the algorithm wrappers out of main.py and into separate files 
(see the ann_benchmarks/algorithms/ folder)
* Added support for datasets containing different kinds of data point 
(see ann_benchmarks/data.py)
* Added support for algorithm implementations running in external 
processes (see ann_benchmarks/algorithms/subprocess.py), the protocol
directory, and the wrappers provided for annoy-hamming, dolphinn, and MIH in
the install directory
* Broke up the results files and created a directory hierarchy for them 
(see ann_benchmarks/results.py)
* Made the results files store the complete results rather than 
precomputed quality metrics
* Added an interactive HTML5 plot generator (see createwebsite.py)
* Added new datasets (see install/data-*.sh)
* Added new quality metrics (see ann_benchmarks/plotting/results.py)
* Added new algorithms (see install/lib-*.sh)

The only core change that we have done is how we handle the _ThreadPool_ in case of running the benchmark without the `--single` flag that forces single-core, single-thread. In our code, the query time of each query is recorded, while the the current ann-benchmarks records the time needed to answer all the queries "once". Our behavior is that"multi-threading" makes the experiments run faster, but the results we get are (almost) identical to when you run in a single-threaded mode. We also added support for a *batch mode*, in which the algorithm takes care of handling all the queries and might use multi-threading, up for it to decide. The batch mode was in particular needed for the FAISS GPU integration.

The following image shows a comparison between single-core/single-thread (**annoy**), threadpool with a single time measurement (**annoy-pool**), and threadpool with individual query time measurements (**annoy-pool-query-based**). We clearly see a difference between the single/individual run, but they keep the actual performance shape for the algorithm. The threadpool with a single measurement at the beginning and end seems to have a lot of overhead for the management included, only at very high recall we see a speedup that we would expect from parallelizing queries.
screen shot 2017-10-17 at 21 29 58

Looking forward to your thoughts!


erikbern:
this is pretty big :)

i noticed a lot of the code is contained in `protocol/` and `random-input-generation/` – these are third party libraries right? if so, let's figure out a way to not include them. for instance you could clone them from github when installing.

i also saw some 3rd party JS libraries that ideally shouldn't be included – usually many JS libraries can be run directly from a CDN, or you could have a simple package.json to set it up using npm.

would be great to reduce the size of this PR down from ~70k lines to something more manageable like a few thousand at most – that's what I expect the "real" size of this PR to be

thanks for the contributions!

erikbern:
Another thing – is there a reason why annoy-hamming is a separate algorithm? Can you merge those change back into Annoy instead?

maumueller:
> This is great – down from 70kloc to 20kloc already :)

Well, thanks for your pointers that greatly improve this PR. :-)

> More questions:

> What's install/lib-annoy.cpp? Why can't you use the Python version?
> Another thing – is there a reason why annoy-hamming is a separate algorithm? Can you merge those changes back into Annoy instead?

I could merge annoy-hamming into annoy, but I think I would need some help with the python wrapper. `install/lib-annoy.cpp` will be removed.

> Why tools/split-dataset.rb in Ruby? I'm not a big fan of mixing too many languages – there's already Python, C++, maybe some Java (?) in there, bash scripts...

Sure, I will use a python equivalent from sklearn. 

> Great if you can remove the Chart library from this PR. You can include it from Cloudflare: > https://cdnjs.com/libraries/Chart.js/1.0.2

I need a patched version of Chart.js. In the worst case, I can just apply the patch manually when `install.sh` is running. Will think about that one.  

> Same with Bootstrap

Will do that.

maumueller:
I've removed the annoy wrapper, removed the split-dataset ruby script (duplicated functionality on top of being in ruby), moved most of the website setup to CDN (only chart.js is installed and patched locally, but without keeping a copy). @ale-f, can you comment on `protocol`?

erikbern:
Great progress! 5kloc now :)

What is the patch you need in Chart.js? Maybe you could modify the DOM or the source code from outside the library

If you create a PR for Annoy featuring Hamming distance, I can take a look at it and figure out a way to make it work. I think it's relatively straightforward

ann_benchmarks/algorithms/constructors.py
```diff
@@ -0,0 +1,63 @@
+from __future__ import absolute_import
+import sys
+import importlib
+import traceback
+
+__constructors = [
```
erikbern: could probably just import these rather than using strings?

ann_benchmarks/algorithms/constructors.py
```diff
@@ -0,0 +1,63 @@
...
+            assert hasattr(module, symbol), """\
+import error: module %s does not define symbol %s""" % (name, symbol)
+            available_constructors[symbol] = getattr(module, symbol)
+    except ImportError:
+        try:
+            t, v, tb = sys.exc_info()
+            traceback.print_exception(t, v, tb)
+        finally:
+            del tb
+        print """\
```
erikbern: should try to make this work with python 3 as well. mostly just make sure print statements have parentheses

ann_benchmarks/data.py
```diff
@@ -0,0 +1,36 @@
+from __future__ import absolute_import
```
erikbern: what is `data.py` used for?

ann_benchmarks/results.py
```diff
@@ -0,0 +1,134 @@
...
+
+                        for a in algos:
+                            adir = os.path.join(qdir, "algo=%s" % a)
+                            runs = _next_part(adir, None, "")
+
+                            if not runs:
+                                continue
+                            else:
+                                for r in runs:
+                                    rpath = os.path.join(adir, r)
```
erikbern: not sure what's going on here but it seems a bit crazy to have 6 for-loops nested like this? seems like it could just be a series of loops with no nesting

install/data-fashion-mnist.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
```
erikbern: should probably also make all the datasets separate docker containers that download and preprocess them... i can give it a shot later (not in this PR)

install/website-chart.js.2.5.0.patch
```diff
@@ -0,0 +1,51 @@
+--- Chart.js	2017-05-25 09:56:43.000000000 +0200
++++ website/js/Chart.js	2017-10-18 09:22:45.000000000 +0200
+@@ -8868,6 +8868,7 @@
+ 			beforeLabel: helpers.noop,
+ 			label: function(tooltipItem, data) {
+ 				var datasetLabel = data.datasets[tooltipItem.datasetIndex].label || '';
++				var customLabel = data.datasets[tooltipItem.datasetIndex].data[tooltipItem.index].label || '';
```
erikbern: rather than patching this library, i think you can just override the `label` function from outside.. a bit hacky but slightly less hacky than the patching

install/website-chart.js.2.5.0.patch
```diff
@@ -0,0 +1,51 @@
...
++				var customLabel = data.datasets[tooltipItem.datasetIndex].data[tooltipItem.index].label || ''; 
+ 				return datasetLabel + ': ' + tooltipItem.yLabel;
+ 			},
+ 			labelColor: function(tooltipItem, chartInstance) {
+@@ -8886,7 +8887,11 @@
+ 
+ 			// Args are: (tooltipItems, data)
+ 			beforeFooter: helpers.noop,
+-			footer: helpers.noop,
++			footer: function(tooltipItem, data) {
```
erikbern: same as above... this in particular looks very amenable to just installing the function yourself – that's probably the library's intention

ann_benchmarks/algorithms/constructors.py
erikbern: (not blocking for PR)

ann_benchmarks/algorithms/constructors.py
erikbern: (also not blocking)

ann_benchmarks/algorithms/constructors.py
ale-f: This code has two goals:

* we want to expose only some explicitly enumerated constructors through the configuration file; and
* we want to be able to distinguish between a constructor that's not been installed -- in which case we should skip its experiments --  and one that's not known to the framework (which should be treated as a hard error).

I suppose this *could* also be done by putting every constructor symbol in a separate `import` statement wrapped up in its own `try`/`except ImportError` block, but... I don't know, that seems messier to me.

ann_benchmarks/data.py
ale-f: It's part of the support code for datasets of different types. `data.py` defines functions for serialising and unserialising each entry in a dataset, and the loader in `ann_benchmarks.datasets` uses its `type_info` map to find these functions. (`ann_benchmarks.algorithms.subprocess`, our mechanism for talking to experiments running in external processes, also uses them to convert entries back into plaintext for transmission.)

ale-f:
The `protocol/` folder contains sample implementations of a text-based protocol that we designed for communicating with experiments running in external processes. (The `ann_benchmarks.algorithms.subprocess` module implements the framework's half of the protocol.)

I agree that we might want to move it somewhere else (a new repository in the upcoming organisation, perhaps?), but it's not really third-party code.

erikbern:
ok – i'm not convinced the `protocol/` stuff is needed and the code is highly confusing so it would be great if we can remove the dependency on it. why do you need a data serialization format in the first place? and even if you do, why not use one of the many readily available ones such as protobuf?

ann_benchmarks/data.py
erikbern: ok is this the same data format as `protocol`?

protocol/java/src/dk/itu/epbprtv0/Runner.java
```diff
@@ -0,0 +1,133 @@
+package dk.itu.epbprtv0;
```
erikbern: What's the purpose of all the Java code? I don't think any of the ANN algorithms are written in Java right?

Would like to keep this to Python and C++ – is there a way to remove the Java code?

ale-f:
The `protocol/` stuff is supposed to be used to test algorithms that don't already have, or that can't have, Python wrappers (I think @maumueller's been using it for a Java set similarity implementation, for example). I think that's a useful thing to be able to support. The directory contains self-contained implementations of that protocol, along with some documentation; we might want to improve that documentation, and to move the actual code to another repository, but it's worth keeping in some form.

The code in `ann_benchmarks.data` just parses and unparses the plain text format that the datasets already use; I don't think that's a lot of extra complexity.

erikbern:
I'm happy to merge this provided that
* We remove the Chart.js from the PR - think it should be easy to monkeypatch the CDN version by just overloading functions
* We remove all of the stuff in `protocol/` – especially the Java code. I think it's unfortunate to depend on Java to run this (makes installing it much more complex). I also don't see the value of using a highly custom serialization format as opposed to something off the shelf (I mentioned protobuf, but there's also HDF5 and other ones). To be clear even if the code was factored out into its own repository, I'd be very hesitant to use it.

Looking forward to hear your thoughts.

ale-f:
I have a feeling we may be talking at cross purposes here...

The protocol we use for communicating with experiments in external processes *is* a plain-text protocol. The code in `ann_benchmarks.data` was already responsible for loading the existing plain-text entries from datasets; it can now also recreate those plain-text representations from the loaded data points.

None of the code in `protocol/` is a dependency of the framework: that directory contains sample implementations of the *algorithm* side of the plain-text protocol that implementations could either adapt or use directly in order to communicate using the protocol. The corresponding (Python) code in the framework, in `ann_benchmarks.algorithms.subprocess`, implements the `fit(X)` and `query(q, k)` methods by sending and receiving protocol messages; the implementation language on the other end is of no consequence.

erikbern:
I'm not quite following but it sounds like the `protocol/` stuff could be removed then? If it's not a dependency of the framework.

yurymalkov:
@maumueller Hi! Thanks for the update.
I wonder, how did you solve the problem with the mixed singlethread/multithread algorithm tests?

As it follows from your plot there is a difference between the multi-threaded and single-thread performance (the difference, I think, should be more pronounced for datasets/algorithms with higher QPS).

maumueller:
Some more reasoning for the protocol/subprocess:

At the moment, the `protocol` is used during compilation for `install/lib-mih.sh` and `install/lib-dolphinn.sh`. In principle, the protocol and wrappers could also be moved into these libraries, but there still needs to be a place to maintain them. (And it would be much harder to maintain them decentrally.)

In principle, the subprocess with this plain-text protocol is supposed to be a more general and potentially easier way to include algorithms. We did this for the algorithms above, and it is currently used by the FALCONN people since their python wrappers are broken and they still want to test performance. I’m also talking with the developers of https://elki-project.github.io to include their implementations into ann-benchmarks and they will use the Java protocol. Sure, Java will hurt performance, but I don’t see why there shouldn’t be a way for them to benchmark their implementations.

@erikbern Do you feel that you understand the purpose now?

maumueller:
@yurymalkov I think there is no **solution** when using the ThreadPool. In my mind, the solution is that ann-benchmarks is a single process, but it has two query modes: single_query and batch_query. The implementation should decide how it wants to handle parallizing queries in “batched mode”.  (The standard implementation could for example be that it uses a ThreadPool to parallize single queries.)

The multi-threading behavior in this PR measures each query independently and as can be seen from the plot is penalized for using the ThreadPool. However, the performance plot is basically the single plot with a small offset (note that QPS is already pretty high), so I think it shows the individual query behavior well. The experiment run 4x faster through, though.  I’d love to hear whether we could come up with a better way!

erikbern:
ok i'm still somewhat adamant about not merging the protocol stuff since the code looks super kludgy and i don't understand why we couldn't have used an off the shelf serialization library. i mean just the name of it is `epbprtv0` – it's like my cat walked over the keyboard :)

as a compromise, can you remove lib-mih and lib-dolphinn from this pull request for now? we can merge them as a separate pull request and defer that discussion until later.

i'd rather get this in and start doing my own cleanup first – in particular i think dockerizing each algorithm individually will be a lot clearer, and add a lot of structure. once i'm done with that, i think it will be easier to figure out a clean way to support java. but either way i think it could/should be a separate discussion

oh and can you also address my comment about `Chart.js` and `result.py` – those are my only remaining outstanding concerns besides the protocol/java stuff

erikbern:
sorry by the way if i'm sounding grumpy here. just wanted to mention again that i'm very excited about merging this in! also very excited about the prospect of having more people working on this project.

maumueller:
No problem at all, great to get feedback! 

I think it's fair enough to separate the subprocess/protocol from this PR since it really is a separate issue. 

I'm having a look `Chart.js` at the moment. What do you mean with `result.py`? I didn't see a reference to that within this conversation.

maumueller:
I removed the subprocess/protocol stuff and monkey patched the changes to `Chart.js`.

erikbern:
ok excellent – will take another look later today.

i meant `results.js`. There's some code there that features 6 nested for loops. Seems a bit weird.

maumueller:
Alec changed the nested for loop monster with the latest commit.

erikbern:
awesome – thanks

erikbern:
I guess this is ready to be merged... hoping to have time to review this weekend

erikbern:
Now that Hamming might be working in Annoy, I want to get back to this PR this week.

One more question – how do I run this? What documentation should I follow? Would be great to have a list of commands I need to run

maumueller:
We gave a very basic documentation in the README.rst. We can update the documentation if you think it's not sufficient.

Verbatim:

Experiment Setup
================

Running a set of algorithms with specific parameters works as follows:

* Check that ``algos.yaml`` contains the parameter settings that you want to test
* To run experiments on SIFT, invoke ``python -m ann_benchmarks/main --dataset sift-data --query-dataset sift-query --distance euclidean``. See ``python -m ann_benchmarks/main --help`` for more information on possible settings. Note that experiments can take a long time. 
* To process the results, either use ``python plot.py`` or ``python createwebsite.py``. An example call: ``python createwebsite.py --plottype recall/time --latex --scatter --outputdir website/``.

erikbern:
thanks! sorry didn't see this among all the other code changes

erikbern:
so running `ann_benchmarks/main` will download the datasets, install all algorithms etc?

maumueller:
No, installing algorithms and data sets is done by invoking `install.sh` in the main directory (as discussed in the README). However, this only installs SIFT, others have to be installed manually.

erikbern:
Ok, sounds good.

How do you feel about storing all the data sets in preprocessed form in S3? I already store a few datasets there – could add the rest as well. Think it would be useful to store the train/test split and the correct answers as well (as found by exhaustive search).

maumueller:
That would make the whole installing process much faster, good idea.

README.rst
```diff
@@ -1,93 +1,120 @@
...
-
 Benchmarking nearest neighbors
-------------------------------
+==============================
+
+This project contains some tools to benchmark various implementations of approximate nearest neighbor (ANN) search for different metrics. Visit
+http://sss.projects.itu.dk/ann-benchmarks to see the results of this benchmark.
 
-This project contains some tools to benchmark various implementations of approximate nearest neighbor (ANN) search.
+The predecessor can be found at https://github.com/erikbern/ann-benchmarks.
```
erikbern: note to myself: should prob remove this after merge

erikbern:
boom

will do a bunch of post-merge cleanup

- [ ] make CI work again, possibly by dockerizing each algo
- [ ] preprocess all datasets and store in S3

PR #42, opened 2017-11-23
NYtimes dataset

maumueller:
This pull request adds support for the NYtimes dataset.  The NYTimes bag-of-words dataset is transformed into a 256 dimensional real-valued value by normalizing the word occurrences matrix using tfidf and then projecting done to 256 dimensions.

The bag-of-words transformation is kept general, since it potentially generalizes to other datasets.


erikbern:
thanks! i'll run it and upload the resulting dataset

PR #52, opened 2018-01-21
Nearest neighbors for training set

dataplayer12:
Hi, this is in response to the issue I raised.

### DIFF
--- a/ann_benchmarks/datasets.py
+++ b/ann_benchmarks/datasets.py
@@ -33,6 +33,35 @@ def get_dataset(which):
 # Everything below this line is related to creating datasets
 # You probably never need to do this at home, just rely on the prepared datasets at http://vectors.erikbern.com
 
+def write_helper(bf,data,count,nn_dataset,d_dataset,remove_self=False):
+    """
+    bf: instance of BruteForceBLAS object
+    data: data to write NNs for
+    count: number of nearest neighbors to query
+    nn_dataset: dataset to store indices of NNs
+    d_dataset:dataset to store distances
+    remove_self: bool (if True will remove self index from results)
+    For train data, the NNs will include the datapoint itself as well
+    """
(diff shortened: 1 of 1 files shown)


erikbern:
Nice. Can you fix two things?

1. You just renamed the datasets in HDF5. ann-benchmarks uses those names in other places so you'd have to rename them elsewhere too. Or just keep the name of the current datasets and name so that it's `distances` and `train_distances` (a bit inconsistent maybe, but seems fine)
2. There's a bit of code redundancy between generating the train neighbors and test neighbors. Can you put it in a for loop or use a function or something to reduce the code duplication?

erikbern:
I also wonder how much this will inflate the size of the indexes. I'll take a look. Maybe not a big issue.

dataplayer12:
I have added a function to avoid redundancy. I think it would be okay if we skip storing distances for train data. Since the indices of 100 neighbors are provided, the time to compute distances of NNs is still reduced by a factor of 10^4.

erikbern:
Looks good. Was thinking actually what could make sense is to store the top 100 neighbors and distances for all vectors in the test set, but only 10 for the ones in the train set, just so it takes a bit less memory. What do you think?

dataplayer12:
10 points is too few to calculate commonly used metrics like recall and mAP. If you are concerned about memory, I can enable chunking in the h5 file so that only a small chunk of the data is read at any time. Currently you store all 1M points in one huge contiguous array. Let me know.

erikbern:
Let's do 100 then, but let's skip the distances.

dataplayer12:
Thanks. In that case, the code should work as is. Look forward to the updated datasets.

erikbern:
Will try to run this at some point. I'm just a bit concerned it will be incredibly slow. It might take days/weeks to run. I'll kick it off and see if it's doable.

PR #53, opened 2018-01-30
Adding dummy algorithms to measure the maximum QPS, enabeling multithreading on FALCONN

yurymalkov:
Added dummy algorithms, which just return np.random.randint(self.len, size=n) as the result, for both multi-threaded and single-threaded tests. 
Should be extremely fast in theory, but in reality I got only 1.37 kQPS int multithreaded tests:

DummyAlgoSingleThread
k-nn: 2e-05
epsilon: 2e-05
largeepsilon: 8e-05
rel: 2.54841
qps: **189150**
build: 0.000294924
candidates: 10

DummyAlgoMultiThread
k-nn: 3e-05
epsilon: 4e-05
largeepsilon: 0.00011
rel: 2.55033
qps: **1366.86**
build: 0.00029707
candidates: 10

Probably np.random.randint does not releases GIL, but IMO it cannot explain 100X difference on a 6-core CPU.

Also changed the mt-flag for FALCONN, as it officially supports multithreading now.


erikbern:
nice – interesting.

why don't we just switch to use multiprocessing instead of multithreading. nmslib is immutable during querying right? since linux is copy-on-write, should be fine memory wise

yurymalkov:
Hm. I am not sure that nmslib is immutable. It uses a shared pool of objects to avoid memory re-allocations.
Anyway we can try testing the perfomance of the dummy algorithm in this case.

erikbern:
ok then the other option is enabling cpu affinity

yurymalkov:
@erikbern Done that. It seems to be working properly, only one particluar core is used according to htop.

requirements.txt
```diff
@@ -6,3 +6,4 @@ numpy==1.13.3
 pyyaml==3.12
 psutil==5.4.2
 scipy==1.0.0
+psutil==5.4.2
```
erikbern: it's alraedy included two lines up :)

ann_benchmarks/runner.py
```diff
@@ -69,8 +70,14 @@ def batch_query(X):
                 pool = multiprocessing.pool.ThreadPool()
                 results = pool.map(single_query, X_test)
             else:
+                p = psutil.Process()
```
erikbern: might as well enable cpu affinity for the thread mode as well.

actually thinking about it it probably makes sense to just drop all other "modes" and only support single-threaded with cpu affinity

requirements.txt
yurymalkov: Haven't noticed it)

ann_benchmarks/runner.py
yurymalkov: I am not sure what the "thread" mode is. 
I agree that the multi-threaded mode does not make much sense (because of the overhead), but the batch mode is probably useful for testing on GPUs.

erikbern: i'm not a big fan of including batch mode since i don't think it's an apples to apples comparison. if some algorithms run in batch mode and some run in single-query mode then it's not a fair comparison.

erikbern:
FYI I'm planning to merge this and re-run the benchmarks in a few weeks. I just published a lot of benchmarks and it took a long time to create them (and like $100 in EC2 costs) so I want to give people a chance to take a look at them first (for instance I have some optimization ideas for Annoy that I might try)

ann_benchmarks/runner.py
yurymalkov: I agree. Probably it is best to present the results for batch mode separately.

yurymalkov:
Ok.

yurymalkov:
@erikbern Are you going to merge this branch?

erikbern:
yeah sorry will take a look at it in the next few days hopefully

do you mind rebasing this? looks like there's a merge conflict right now

erikbern:
thanks!

yurymalkov:
Cool! By the way, can you please notify prior to re-running the tests?

erikbern:
sure, what's the reason? if you need to update something w nmslib i can also just wait for you to let me know once it's down... i'm not in a rush

yurymalkov:
Great, thanks! I'll let you know.

erikbern:
Thanks @yurymalkov 

I'm going to drop the thread pool

yurymalkov:
Cool! Thanks!

PR #55, opened 2018-02-16
Upgrade to use newer NMSLIB api

benfred:
Nice work on all the recent changes! I saw your post this morning and decided play around with this again.

One thing I noticed was that this package is using an old version of the nmslib api. While I kept that version around for backwards compatibility, the newer version has a couple of advantages: you don't
need to call freeIndex, the newer version accepts numpy arrays as inputs etc etc.

Also, I think you are running into this issue https://github.com/searchivarius/nmslib/issues/291 . I noticed that the tests hung here, but when switching the Dockerfile.nmslib to point to my repo of nmslib (with the fix) it worked.

### DIFF
--- a/ann_benchmarks/algorithms/nmslib.py
+++ b/ann_benchmarks/algorithms/nmslib.py
@@ -4,10 +4,12 @@ import nmslib
 from ann_benchmarks.constants import INDEX_DIR
 from ann_benchmarks.algorithms.base import BaseANN
 
+
 class NmslibReuseIndex(BaseANN):
     @staticmethod
     def encode(d):
         return ["%s=%s" % (a, b) for (a, b) in d.iteritems()]
+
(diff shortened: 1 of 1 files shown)


erikbern:
cool thanks, will merge if this builds!

erikbern:
btw fun to see a PR from you – have been following your blog for a while and seems like you're working on some fun stuff. I know at least one person have confused the two of us due to similar names :)

benfred:
Thanks! I've been following your blog for a while too =).

I think the build won't work until that issue in nmslib is merged. It looks like the build is hanging on the deadlock in pybind11. I'll restart the CI here once the fix is merged into master in nmslib.

erikbern:
got it, thanks

erikbern:
thanks!

PR #54, opened 2018-02-06
Various updates to plotting

erikbern:
Made it a bit faster, and now using consistent line styles for all algorithms.

Also including results for a bunch of datasets

PR #57, opened 2018-02-19
Bring back create_website.py

maumueller:
This PR adds back support for running `createwebsite.py`. This script builds a website containing interactive plots for all results that are stored. Enabling support for this resulted in some questionable choices after #54. In particular, 

- Needed to add auxiliary helpers to load all the results. It's not a lot of code, but it seems hard to maintain with other parts of code such as `get_result_filename`.
- Needed to convert a lot of data, because the plotting utils now always assume an `x` and `y` value; however, generating all the plots uses a dict from which we can read of the `x` and `y` values. This is really going to hurt performance of the script for large result repositories.

In general, `create_website.py` is a monolithic beast because of all the html output that is stored directly in it. I'd welcome comments on how to improve it (using some template engine?).

On the way, I decoupled some details of the metric computation from the file format that is now used. That made it possible to provide some unit tests for the metric computation, that I felt should be in place also for documentation reasons. 

In general, since the change to the hdf5 file format, plotting became _horribly slow_. It's about 20 times slower for similar result files than the previous version that stored everything in JSON and used a different format. I feel the need that we really have to make this part more efficient.

### DIFF
--- a/ann_benchmarks/results.py
+++ b/ann_benchmarks/results.py
@@ -30,3 +30,27 @@ def load_results(dataset, count, definitions):
             f = h5py.File(fn)
             yield definition, f
             f.close()
+
+def _get_leaf_paths(path):
+    if os.path.isdir(path):
+        for fragment in os.listdir(path):
+            for i in _get_leaf_paths(os.path.join(path, fragment)):
+                yield i
+    elif os.path.isfile(path):
+        yield path
+
+def load_all_results():
+    import re
+    #TODO(Martin) Must be adaptive to get_result_filename
+    regex = r"results\/(?P<dataset>[^\/]+)\/(?P<count>\d+)\/(?P<algorithm>[^\/]+)\/(?P<distance>[^_]+)_(?P<params>.*)"
+    for fn in _get_leaf_paths("results/"):
+        match = re.match(regex, fn)
+        if match:
+            d = {}
+            for k in ["count", "dataset", "distance", "algorithm"]:
+                d[k] = match.group(k)
+            d["count"] = int(d["count"])
+            f = h5py.File(fn)
+            yield d, f
+            f.close()
+
--- a/ann_benchmarks/results.py
+++ b/ann_benchmarks/results.py
@@ -41,16 +41,11 @@ def _get_leaf_paths(path):
 
 def load_all_results():
     import re
-    #TODO(Martin) Must be adaptive to get_result_filename
-    regex = r"results\/(?P<dataset>[^\/]+)\/(?P<count>\d+)\/(?P<algorithm>[^\/]+)\/(?P<distance>[^_]+)_(?P<params>.*)"
     for fn in _get_leaf_paths("results/"):
-        match = re.match(regex, fn)
-        if match:
-            d = {}
-            for k in ["count", "dataset", "distance", "algorithm"]:
-                d[k] = match.group(k)
-            d["count"] = int(d["count"])
+        try:
             f = h5py.File(fn)
-            yield d, f
+            yield f
             f.close()
+        except:
+            pass
 
--- a/ann_benchmarks/results.py
+++ b/ann_benchmarks/results.py
@@ -31,21 +31,13 @@ def load_results(dataset, count, definitions):
             yield definition, f
             f.close()
 
-def _get_leaf_paths(path):
-    if os.path.isdir(path):
-        for fragment in os.listdir(path):
-            for i in _get_leaf_paths(os.path.join(path, fragment)):
-                yield i
-    elif os.path.isfile(path):
-        yield path
-
 def load_all_results():
-    import re
-    for fn in _get_leaf_paths("results/"):
-        try:
-            f = h5py.File(fn)
-            yield f
-            f.close()
-        except:
-            pass
+    for root, _, files in os.walk("results/"):
+        for fn in files:
+            try:
+                f = h5py.File(os.path.join(root, fn))
+                yield f
+                f.close()
+            except:
+                pass
 
--- a/create_website.py
+++ b/create_website.py
@@ -186,7 +186,7 @@ def prepare_data(data, xn, yn):
         res.append((algo, algo_name, result[xn], result[yn]))
     return res
 
-def get_latex_plot(all_data, xm, ym, plottype):
+def get_latex_plot(all_data, xn, yn, xm, ym, plottype):
     latex_str = """
 \\begin{figure}
     \\centering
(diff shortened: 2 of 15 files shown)


erikbern:
Thanks for fixing this. I'm happy to reconsider any decisions made in #54 if it made this harder.

Think honestly generating nice HTML output could potentially supersede the matplotlib plots. I'd rather have nice interactive vector graphics using some JS library

> Needed to add auxiliary helpers to load all the results. It's not a lot of code, but it seems hard to maintain with other parts of code such as get_result_filename.

I have an idea of how to make this simpler (think you can just use os.path.walk and then read the algorithm etc from the HDF5 metadata)

> Needed to convert a lot of data, because the plotting utils now always assume an x and y value; however, generating all the plots uses a dict from which we can read of the x and y values. This is really going to hurt performance of the script for large result repositories.

We can fix this – the reason I made that change was that I thought it was unnecessary to compute *all* metrics during plotting even though typically only 1 is used (well, 2, but the qps is trivial).

> In general, create_website.py is a monolithic beast because of all the html output that is stored directly in it. I'd welcome comments on how to improve it (using some template engine?).

jinja2 is the most popular templating library in Python, would consider using it.

> 
> On the way, I decoupled some details of the metric computation from the file format that is now used. That made it possible to provide some unit tests for the metric computation, that I felt should be in place also for documentation reasons.

Can you add something to `.travis.yml` to make sure those tests run as a part of CI?

> 
> In general, since the change to the hdf5 file format, plotting became horribly slow. It's about 20 times slower for similar result files than the previous version that stored everything in JSON and used a different format. I feel the need that we really have to make this part more efficient.

This is really weird. HDF5 should be extremely fast. Are you sure this is related to the file format? I wonder if maybe I increased the number of data points in the test set or something? HDF5 would be pretty far down my list of suspects

maumueller:
> Think honestly generating nice HTML output could potentially supersede the matplotlib plots. I'd rather have nice interactive vector graphics using some JS library

I think they can coexist. Currently, the website still uses the plotting engine to create a summary plot that is identical to with the plots we have in `README.md` at the moment. 

>> Needed to convert a lot of data, because the plotting utils now always assume an x and y value; however, generating all the plots uses a dict from which we can read of the x and y values. This is really going to hurt performance of the script for large result repositories.

> We can fix this – the reason I made that change was that I thought it was unnecessary to compute all metrics during plotting even though typically only 1 is used (well, 2, but the qps is trivial).

Yes, it's a pretty particular scenario when building the website. I think we should make it a bit more generic again (storing the result in a dict), but from `createwebsite.py` we pass a `metrics.keys()` as the metrics we want to compute, from `plot.py` we pass `['k-nn', 'qps']`.

>> In general, create_website.py is a monolithic beast because of all the html output that is stored directly in it. I'd welcome comments on how to improve it (using some template engine?).

> jinja2 is the most popular templating library in Python, would consider using it

Cool, will try (probably over the weekend) before this should be merged.

 >> On the way, I decoupled some details of the metric computation from the file format that is now used. That made it possible to provide some unit tests for the metric computation, that I felt should be in place also for documentation reasons.

> Can you add something to .travis.yml to make sure those tests run as a part of CI?

Done!

>> In general, since the change to the hdf5 file format, plotting became horribly slow. It's about 20 times slower for similar result files than the previous version that stored everything in JSON and used a different format. I feel the need that we really have to make this part more efficient.

> This is really weird. HDF5 should be extremely fast. Are you sure this is related to the file format? I wonder if maybe I increased the number of data points in the test set or something? HDF5 would be pretty far down my list of suspects.

I have honestly no idea at the moment why it became so slow. The format was more succinct previously; if I remember correctly, for each query we had just stored the distance of the true k-th nearest neighbor and didn't handle lists of a hundred distance values in the metric computation.

maumueller:
Some updates:

- I refactored major parts of `createwebsite.py`. It's still not using a templating engine, but it's much easier to understand now. 
- I don't read the attributes off the path to the result file, but just return all the files found in the `results/` directory and read the information from the actual file. However, most of the information needed wasn't stored in the file, so I had to add it.

With regard to plotting performance:
- The culprit is the hdf5 file format. The current code treats access to the file essentially as being for free, but according to my measurements (on an SSD) it comes with a high penalty. E.g., accessing the true nearest neighbors and the queries comes with a penalty of roughly 100ms each. Before the hdf5 change, computing the recall took .024s for 100-NN, now it's .26s. I would suggest that we essentially put the file content into memory before computing the metrics. (Just doing `list(f['neighbors'])` where `f` is an hdf5 file.)

One problem a ran into: 
- For some algorithms, reading the stored files results in strings being represented as bytes (e.g., `b'faiss-ivf'`). I currently included a workaround for this in `create_website.py`. Do you know what is going on? (Running annoy and faiss-ivf on random-xs-20-angular results in this problem for faiss, but not for annoy on my machine.)

erikbern:
> I don't read the attributes off the path to the result file, but just return all the files found in the results/ directory and read the information from the actual file. However, most of the information needed wasn't stored in the file, so I had to add it.

this is great!

> 
> The culprit is the hdf5 file format. The current code treats access to the file essentially as being for free, but according to my measurements (on an SSD) it comes with a high penalty. E.g., accessing the true nearest neighbors and the queries comes with a penalty of roughly 100ms each. Before the hdf5 change, computing the recall took .024s for 100-NN, now it's .26s. I would suggest that we essentially put the file content into memory before computing the metrics. (Just doing list(f['neighbors']) where f is an hdf5 file.)

Ok – if the remedy is as easy as just converting it to a list, let's just do that!

Super weird that HDF5 is so slow, I'm surprised by it

> One problem a ran into:
> 
> For some algorithms, reading the stored files results in strings being represented as bytes (e.g., b'faiss-ivf'). I currently included a workaround for this in create_website.py. Do you know what is going on? (Running annoy and faiss-ivf on random-xs-20-angular results in this problem for faiss, but not for annoy on my machine.)

I'm fairly sure this is because some algorithms run in Python 2, some in Python 3. I noticed some inconsistencies with bytes/strings too but it didn't have an impact on the benchmarks so I just ignored it... shouldn't be very hard to fix if it's causing issues

ann_benchmarks/results.py
```diff
@@ -30,3 +30,22 @@ def load_results(dataset, count, definitions):
             f = h5py.File(fn)
             yield definition, f
             f.close()
+
+def _get_leaf_paths(path):
```
erikbern: think you can just use os.walk here and make this a few lines shorter (don't have to implement the recursion yourself)

install.py
```diff
@@ -9,7 +9,10 @@
 
 def build(library):
     print('Building %s...' % library)
-    client.images.build(path=os.getcwd(), tag='ann-benchmarks-%s' % library, rm=True, dockerfile='install/Dockerfile.%s' % library)
+    try:
+        client.images.build(path=os.getcwd(), tag='ann-benchmarks-%s' % library, rm=True, dockerfile='install/Dockerfile.%s' % library)
+    except docker.errors.BuildError as err:
+        print("Build error: {0}".format(err))
```
erikbern: what's the rationale for continuing on errors? don't have strong feelings, but seems unrelated to the rest of the changes

ann_benchmarks/results.py
maumueller: Done. Thanks for the suggestions.

install.py
maumueller: Ah, I've added it after (for some reason) the faiss container couldn't be build, and everything following faiss couldn't be installed as a result. (They've released a new version, maybe I pulled somewhere in between where it didn't compile.) 

Since we don't have control over the repos that we are using, it's likely that once in a while a container cannot be build. It shouldn't stop all the other containers from getting installed, should it?

create_website.py
```diff
@@ -275,232 +312,138 @@ def create_plot(ds, all_data, xn, yn, linestyle, additional_label = "", plottype
...
+
+    with open(args.outputdir + "index.html", "w") as text_file:
+        text_file.write(output_str)
+
+def load_all_results():
+    """Read all result files and compute all metrics"""
+    all_runs_by_dataset = {}
+    all_runs_by_algorithm = {}
+    for f in results.load_all_results():
+        properties = dict(f.attrs)
```
maumueller: >> One problem a ran into:

>> For some algorithms, reading the stored files results in strings being represented as bytes (e.g., b'faiss-ivf'). I currently included a workaround for this in create_website.py. Do you know what is going on? (Running annoy and faiss-ivf on random-xs-20-angular results in this problem for faiss, but not for annoy on my machine.)

> I'm fairly sure this is because some algorithms run in Python 2, some in Python 3. I noticed some inconsistencies with bytes/strings too but it didn't have an impact on the benchmarks so I just ignored it... shouldn't be very hard to fix if it's causing issues

That should be it. Faiss still runs on Python 2. I checked some h5py resources, but couldn't find a solution. So, I'm converting the `f.attrs` dictionary myself. (I'm building strings from these descriptions, but I cannot do it when they are bytes.)

maumueller:
Ok, I've updated the code to cache all accesses to the hdf5. This made `create_websites.py` around 5x times faster on a small set of result files.

erikbern:
this is really nice – let me know when it's ready to be merged

requirements.txt
```diff
@@ -6,3 +6,5 @@ numpy==1.13.3
 pyyaml==3.12
 psutil==5.4.2
 scipy==1.0.0
+scikit-learn==0.19.1
+jinja2=2.10
```
erikbern: this should be `==` – travis is failing on it right now

maumueller:
Ok, I think I'm quite happy with the result now. Feel free to merge.

erikbern:
Great, let me make sure I can run locally and then I'll merge.

Do you have any thoughts on where to host this? I have http://vectors.erikbern.com set up as an S3 bucket. I could register another domain also (eg ann-benchmarks.com is $9)

maumueller:
For the paper we used sss.projects.itu.dk/ann-benchmarks. We could host it there again. (However, I cannot give you access to it.) I wouldn't get as far renting a domain for it. Just something that we can point to from the repo here should suffice, shouldn't it?

erikbern:
The money is negligible here, so I'd have a marginal preference for setting up a separate domain for this. Partly because it's a nicer URL, partly because that way we can set it up so multiple people can upload. It's trivial to set up in AWS (just need to add a route53 entry for it, and point it to the S3 bucket). I can do that, and also set up AWS keys so that you can upload to it.

maumueller:
Sounds excellent.

install.py
erikbern: that's a valid point, makes sense

erikbern:
Thanks @maumueller !

Haven't had time to give this a thorough review, but let's merge this and I'll take a look at it as a part of re-running the benchmarks (which I'm planning to do in the near future)

PR #63, opened 2018-03-05
removing FALCONN for now (since the Python wrapper is obsolete)

ilyaraz:
For now, the Python wrapper of FALCONN is seriously obsolete, thus the results are outdated. Will try to work on it in the not so distant future to add it back.


erikbern:
Thanks @ilyaraz 

If you can send an updated wrapper for FALCONN in the next few weeks, I'll re-run the benchmarks with it.

ilyaraz:
I will try to, though can't really commit to anything (there is a major conference deadline on April 6th). But definitely, at some point...

erikbern:
ok no worries, i can drag it out a bit

Issue #41, opened 2017-11-21
Split up building and query parameters


maumueller:
Currently, we spend a lot of time basically building the same data structure over and over again. What we could do instead, is split up _build_ and _query_ parameters. For example, Annoy is currently configured by

```
    annoy:
      constructor: Annoy
      base-args: ["@metric"]
      run-groups:
        annoy:
          args: [[100, 200, 400], [100, 200, 400, 1000, 2000, 4000, 10000,
                20000, 40000, 100000, 200000, 400000]] # number of trees x number of points to gather
```

It would save a lot of time, if one were to separate query and build parameters in the following way: 
```
    annoy:
      constructor: Annoy
      base-args: ["@metric"]
      run-groups:
        annoy:
          build-args: [[100, 200, 400]] # number of trees 
          query-args: [[100, 200, 400, 1000, 2000, 4000, 10000,
                20000, 40000, 100000, 200000, 400000]] # number of points to gather during query
```
The idea would now be to build the data structure for each `build-args` group, and query it for all (combinations of) `query-args`. This should be rather straight-forward to do.

Should be labeled as "enhancement".

erikbern:
i like the idea – actually thought about something similar too. would help with the slow nmslib index building time as well

maumueller:
This has been added, closing.

PR #65, opened 2018-03-08
Add support for reusing built data structures

ale-f:
This commit adds support for reconfiguring an already-built data structure with new query parameters, making it possible to reuse one to perform many different experiments. (This was discussed in issue #41.)


ale-f:
Ah, I see why Travis is failing -- the extra metadata that commit 92ea2d93abe4aa1b14c9871395e29b03080a6beb added to the result files got lost in the merge! I'll fix that in a second...

maumueller:
To add some numbers: Running annoy on glove-100-angular took 539m14.031s (current HEAD) to run through, now it takes 72m53.724s on my setup.

erikbern:
I think this feature is great, thanks for implementing.

From what I understand, every docker can now run multiple "definitions"? I'm not sure that's ideal, as it decreases the process isolation.

I thought about how to implement this a few weeks ago, and my idea was to mount an index directory to the Docker container and let each algorithm use that for any intermediate data that can be cached more opportunistically. The upside is the implementation would be a lot simpler, and you would get full process isolation. The downside is that the index building time would be effectively 0 for some runs (any time we would take advantage of the cached index).

maumueller:
I don't immediately see why your proposal would be any simpler, Erik.  It seems that in your proposal each algorithm has to write its index to a file (how many support that? ) and needs custom code to determine whether it can reuse its index or not, which would again boil done to manage on some level which of the part of a definition is build specific and which isn't. This would either mean code equivalent to what Alec is doing or a lot of specific code in each wrapper.

With regard to process isolation. All we do is to call the query method multiple times. In the new setup we make use that some of the parameters are query specific, so these algorithms are actually meant to be used I  this way. If there would be an issue because the process isn't recreated everytime, the same issue should manifest when running say ten times more queries for a single run without changing query parameters.

erikbern:
That's true – I think only Annoy and NMSLIB supports storing indexes on disk. Thanks for the explanation. I think the rationale for building it this way is much stronger.

erikbern:
Btw was looking at the Travis output for Annoy and it's pretty cool – https://travis-ci.org/erikbern/ann-benchmarks/jobs/350785075

ale-f:
@maumueller added support for `faiss-ivf`, so I've added that to the pull request. We also fixed a problem in the query runner: `ThreadPool`'s threads seem to keep running unless the pool is explicitly terminated, so, now that a single query process can create multiple `ThreadPool`s, processes were sometimes running out of threads!

Is there anything else you'd like to see in here?

erikbern:
Sorry, I forgot about this one. Will look at it tonight!

ann_benchmarks/algorithms/faiss.py
```diff
@@ -55,11 +53,17 @@ def fit(self, X):
         index = faiss.IndexIVFFlat(self.quantizer, X.shape[1], self._n_list, faiss.METRIC_L2)
         index.train(X)
         index.add(X)
-        index.nprobe = self._n_probe
         self._index = index
 
+    def set_query_arguments(self, n_probe):
+        self._n_probe = n_probe
```
maumueller: Could anyone of you please correct this typo? It has to be `self.nprobe = n_probe`, as seen above. Otherwise we don't set the probe count for IVF.

erikbern:
Looks like merging #53 caused a conflict with this one. I'll see if I can resolve it

erikbern:
@maumueller @ale-f is this ready to be merged? I can merge #71 or feel free to rebase this one

erikbern:
@maumueller @ale-f ping

ale-f:
OK -- we merged your `master` back into my `build-parameters-1`, so this should be the most up-to-date version of everything. (The CPU affinity changes introduced in #53 break local mode on Mac OS X, which doesn't have the `psutil.Process.cpu_affinity()` function, but we won't put *that* in here as well!)

#71 also contains commit fb6335c89fb6453e17952671dee5aa2c26fee187, which doesn't seem to exist anywhere else. What do you want to do with that?

erikbern:
I'll close #71 and will merge this, OK?

maumueller:
Fine with me 👍

erikbern:
Thanks @ale-f @maumueller !

PR #66, opened 2018-03-08
Fix dataset overview on summary page in create_website.py.

maumueller:
#57 unfortunately had a bug where the dictionary that belongs to a distance measure was added more than once due to an incorrect indentation. This PR fixes that.


erikbern:
thanks @maumueller 

will take a look at this soon

PR #72, opened 2018-04-29
Update PyNNDescent

lmcinnes:
Save off indexes when possible to save time, and update the hyper-parameter selections. If I read the file correctly it should choose 'any' for anything unknown, but override with the specific sets for 'euclidean' and 'angular'?

### DIFF
--- a/algos.yaml
+++ b/algos.yaml
@@ -142,7 +142,7 @@ float:
       base-args: ["@metric"]
       run-groups:
         pynndescent:
-          args: [[5, 10, 15, 20], [4, 8, 16], [40], [1.0, 2.0, 4.0, 8.0]]
+          args: [[10, 20, 40, 80], [4, 8], [30], [1.0, 2.0, 4.0, 8.0]]
   euclidean:
     kgraph:
       docker-tag: ann-benchmarks-kgraph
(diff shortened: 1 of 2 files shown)


erikbern:
@maumueller and @ale-f already implemented some kind of index-reusing mechanism in #65 

Let me merge that first, and I think you can use that one as well

erikbern:
@lmcinnes I've merged #65 now – great if you can update your code to use their new approach.

I'm thinking about re-running the benchmarks in the next week or two, so it would be nice to merge this first

maumueller:
Unrelated to this PR: We should also make sure that the graph-based algorithms from nmslib use the new approach. Is there anyone in particular we could ask to do that, Erik?

Also: Can you make sure to store the results of the experiments, Erik? We need to make a journal version of our paper and I’d like to include the newest results.

erikbern:
I think @searchivarius is the right person for nmslib. The "old approach" will work though, it will just be a bit slower.

I can store the results of the experiments. Could potentially just upload them to S3 or something similar.

searchivarius:
@maumueller I am a bit out of context (glossing over the comments didn't help). What's the new approach? Thanks!

erikbern:
@searchivarius sorry for the lack of context. See #65 – ann-benchmarks now supports reusing the datastructure for different query parameters

lmcinnes:
@maumueller I think I read the changes correctly and hopefully this does the right thing. I would definitely appreciate a check from you to ensure I got this correct.

erikbern:
probably works if the unit tests run successfully :)

lmcinnes:
@maumueller Thanks. I think I finally got the last quirks ironed out, so we should be good to go.

erikbern:
Thanks @lmcinnes 

Will run new benchmarks shortly!

PR #75, opened 2018-05-08
Add MRPT

erikbern:
Turned out to be very simple.

@teemupitkanen any way you can take a quick look at this and see if it makes sense?


erikbern:
This is a quick benchmark vs annoy on random-xs-20-angular:

image

erikbern:
Merging this for now, can always make more changes later.

I'm planning to kick off a new round later this week.

PR #76, opened 2018-05-08
Catch failures for various things a bit more

erikbern:
Noticed FAISS failed building. At least 3 commands didn't care (building the image, running the algos, plotting). See https://travis-ci.org/erikbern/ann-benchmarks/jobs/376353927

This PR makes sure we raise exceptions in all 3 of those places in the case of failures.


erikbern:
issues should be fixed now

PR #77, opened 2018-05-15
Add PANNG(NGT)

masajiro:
We have added our PANNG of NGT library to the benchmarks.
The PANNG is a combination of a graph-based and tree-based indexes.


erikbern:
This is great, thanks!

Can you also add it to `.travis.yml`? That way it will run as a part of the unit tests. Thanks!

erikbern:
Actually nvm why don't I just do that myself

masajiro:
Thank you for merging!
Now I do not need to add PANNG to .travis.yml, do I?

erikbern:
No reason to, see https://github.com/erikbern/ann-benchmarks/pull/78

Issue #47, opened 2017-12-29
10X performance difference between multi-thread and single thread tests


yurymalkov:
Good day!

I have measured the difference in performance when switching the use_threads parameter from True to False on hnsw (nmslib). As it turns out, the difference is really huge, ~ 10X on 1M SIFT for low recalls (see enclosed plots for a 6-core intel cpu), and likely is caused by the python-part overhead of the tester scripts.

Multi-thread (default):
![sift_mt](https://user-images.githubusercontent.com/16265186/34437153-f7afe4fa-ecac-11e7-8f1c-cf382ef9a4b2.png)
Single-thread (use_threads=False):
![sift_st](https://user-images.githubusercontent.com/16265186/34437154-f7cada12-ecac-11e7-9109-a8377da007d1.png)

This makes the current results questionable, especially when compared to FALCONN (which seems to be the only algorithm which takes advantage of not using multi-thread tests). Multithread tests also significantly change the shape of the recall-QPS curve (flat vs rapidly decreasing QPS at low recall).

I suggest using the single thread version for all algorithms. Otherwise, for the case of relatively high QPS (~>2 kps) we just measure the performance of the tester script.

yurymalkov:
Here are the plots for the set of the fastest algorithms on SIFT 1M in two cases:
1. All algorithms are processed in the single thread-regime (by commenting all other options in runner.py)
![sift_st](https://user-images.githubusercontent.com/16265186/34526989-10183c16-f0b5-11e7-93fe-ec5481f97922.png)
2. Default, nothing changed - most of the algorithms are processed in the multi-thread regime.
![sift_mt_old](https://user-images.githubusercontent.com/16265186/34526988-0ff40864-f0b5-11e7-8a79-a94a983415b2.png)

As one can see, the absolute difference in speed is significant and at least three of algorithms have almost the same performance at low recalls in the multi threaded test, which is not adequate to reality.

erikbern:
Hi @yurymalkov – sorry I forgot about this one. This is really weird, not sure what's going on. I'll take a look. @maumueller any idea?

erikbern:
I have a theory what's going on here. Is nmslib/hnsw using threads under the hood, even if queried with a single thread?

[...]

PR #79, opened 2018-05-15
update nmslib for reuse, remove lshkit; add hnswlib; add faiss implementation of hnsw

yurymalkov:


ann_benchmarks/algorithms/faiss_hnsw.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def fit(self, X):
+
+        self.index = faiss.IndexHNSWFlat(len(X[0]),self.method_param["M"])
+        self.index.hnsw.efConstruction = self.method_param["efConstruction"]
+        self.index.verbose = True
+        if(self.metric == 'angular'):
+            X = X / np.linalg.norm(X, axis=1)[:, np.newaxis]
+        self.index.add(X)
+        faiss.omp_set_num_threads(1)
```
erikbern: i thought the cpu affinity code already takes care of this?

yurymalkov: Affinity makes sure that no one takes advantage of multiply threads during the query phase. 
AFAIK openmp and other libs are not aware of the externally set affinity and thus would try to start many threads leading to performance loss.
Faiss uses all cores by default, so probably it is a good idea to set the number of threads to 1.

erikbern: well it's not a fair comparison if some libraries use multiple threads and some libraries don't

Should we remove the CPU affinity and enable an optional thread pool for queries?

yurymalkov: Seems fair to me. 
IMHO having a single-thread regime is obligatory for such libraries.
And if some lib uses multiply threads by default it should be disabled manually in the wrapper. Does not seem like a big deal.

erikbern: Well the problem is when OpenMP/Eigen uses threads under the hood, then it might be hard to fully control how many threads are being used.

I can make it configurable to use multithreading (or multiprocessing) for libraries (Annoy would benefit from this for instance). Would probably enable this by default and opt-out for libraries that implement their own threading.

erikbern: I'm against supporting batch mode since it would make it hard to compare algorithm benchmarks across libraries.

In my real world experience (mostly using Annoy at Spotify), very few use cases are compatible with batch queries

erikbern: > I do not see how threadpool alone will solve this problem. Some algorithms may still create a lot useless threads (in this case num_threads**2).

That's why you make it configurable. So some algorithms will use their own internal threading, so there's no need to use the thread pool (or multiprocessing). Some algorithms do not use threads under the hood so they will benefit from a thread/multiprocessing pool.

I don't have super strong feelings about what the default behavior should be. Could be opt-in or opt-out (either way Annoy would use it and NMSLIB would not, etc)

erikbern:
Thanks, this is great!

I want to kick off a new benchmark very soon (within a week). Only blocker is to figure out the threading stuff.

ann_benchmarks/algorithms/faiss_hnsw.py
yurymalkov: I agree that batch mode is somewhat limited, but there are cases where it is useful.
For instance, it is obviously used in production in Facebook; I know lots of companies use batch mode for kNN on GPUs for processing non-realtime data; K-Means and other clusterization algorithms also may benefit from KNN in batch mode.

I am also a bit worrying about the performance measuring issues, like   https://github.com/erikbern/ann-benchmarks/issues/47 . However this might not be the case for multithreading and now can be monitored by looking at the results of the "dummy" algorithms.

IMO single thread performance is a good common denominator :)

erikbern: i'd like to stick to single thread, but somewhat concerned not all libraries will respect running on a single thread.

but i can run the benchmarks and keep an eye on the cpu usage to make sure they don't use multiple cpus

actually looks like i can do it using docker: https://docs.docker.com/config/containers/resource_constraints/ – i'll try that!

PR #82, opened 2018-05-20
New benchmarks fixes + results

erikbern:
* higher timeout
* fix cosine for MRPT


erikbern:
Still can't get MRPT to work for glove-100-angular. It works consistently if I cap it at 11837 first vectors but fails consistently if I include vector 11837. I don't think it's a coincidence that 11837 * 100 is almost exactly the total size of the matrix.

erikbern:
I uploaded the full result dataset to S3 in the same bucket as the vectors. It's publicly accessible using the aws cli:

```
$ aws s3 ls s3://vectors.erikbern.com/results/2018-06-10/
                           PRE fashion-mnist-784-euclidean/
                           PRE gist-960-euclidean/
                           PRE glove-100-angular/
                           PRE glove-25-angular/
                           PRE nytimes-256-angular/
                           PRE sift-128-euclidean/
2018-06-10 20:56:45     104709 fashion-mnist-784-euclidean.png
2018-06-10 20:56:58     136362 gist-960-euclidean.png
2018-06-10 20:57:09     140171 glove-100-angular.png
2018-06-10 20:57:20     120621 glove-25-angular.png
2018-06-10 20:57:31     135006 nytimes-256-angular.png
2018-06-10 20:57:42     127674 sift-128-euclidean.png
```

erikbern:
@yurymalkov @searchivarius @ale-f @maumueller let me know if you have any thoughts on this, going to write up a blog post shortly

yurymalkov:
@erikbern Cool. Thanks!

maumueller:
Thanks for sharing Erik. I think it’s really useful to give access to the raw data. 

Some points for discussion: 
- Any plans on hosting the result website somewhere? For the current data, it looks like this: https://ktinfty.theoinf.tu-ilmenau.de/~maumueller/website/ I think it’s useful.
- Any reason for not including FAISS-ivf? Sure, on the currently tested datasets it (probably) performs worse than FAISS’ hnsw implementation, but having a completely different technique in the benchmark is worthwhile I think. 
- I think we should really think about more challenging datasets, maybe synthetic ones. There is not much difference on the currently tested datasets between the algorithms. I think we can make differences more pronounced, but it might not be on datasets out there in the wild. 
- I’m astonished by pynndescent’s performance (being written completely in Python)!

Martin

searchivarius:
@maumueller it's not Python, it's Cython, I think.

maumueller:
@searchivarius Thanks for the pointer. I didn’t see the annotations, it uses `numba`‘s just-in-time compilation. Neat.

erikbern:
> Thanks for sharing Erik. I think it’s really useful to give access to the raw data.
> 
> Some points for discussion:
> 
> Any plans on hosting the result website somewhere? For the current data, it looks like this: https://ktinfty.theoinf.tu-ilmenau.de/~maumueller/website/ I think it’s useful.

Thanks for the reminder. I'll set something up. Can register a domain name and serve it out of an S3 bucket.

> Any reason for not including FAISS-ivf? Sure, on the currently tested datasets it (probably) performs worse than FAISS’ hnsw implementation, but having a completely different technique in the benchmark is worthwhile I think.

I think @yurymalkov removed it a while ago. I don't have strong feelings other than being a bit lazy so I'd prefer not to add it right now, but happy to include it next time I run the benchmarks. If you feel super strongly about it I'm open to add it this time.

> I think we should really think about more challenging datasets, maybe synthetic ones. There is not much difference on the currently tested datasets between the algorithms. I think we can make differences more pronounced, but it might not be on datasets out there in the wild.

Yeah I was actually thinking about some other types of datasets. One idea I had was to take something like ResNet, and use the penultimate layer as a vector space embedding of all images in ImageNet. That's one million images times something like 1024 dimensions.

But open to other ideas

yurymalkov:
>Yeah I was actually thinking about some other types of datasets. One idea I had was to take a convolutional neural network, and use the penultimate layer as a vector space embedding of all images in ImageNet (or maybe CIFAR). That's one million images times something like 1024 dimensions (or if you use the full tiny images dataset that's 80 million images) Can use a pretrained model (there's many to download, eg ResNet or VGG or something similar). But open to other ideas.

AFAIK Deep1B (http://sites.skoltech.ru/compvision/noimi/) is made using a similar procedure on billion random images from the web. Another example of such approach (for Amazon product images) is https://github.com/ThomasDelteil/VisualSearch_MXNet . The same can be done for face descriptors. 

>Any reason for not including FAISS-ivf? Sure, on the currently tested datasets it (probably) performs worse than FAISS’ hnsw implementation, but having a completely different technique in the benchmark is worthwhile I think.

I removed it because it was very slow when querying, significantly affecting the total test time. Also, this method falls into a different category - compressed memory indexes for billion scale datasets, so I thought it was not much of interest anyway.

erikbern:
@yurymalkov thanks for the pointers – both look like great resources (the first dataset is easier to use but second is exactly what i meant with my previous comment)

maumueller:
>> Any reason for not including FAISS-ivf? Sure, on the currently tested datasets it (probably) performs worse than FAISS’ hnsw implementation, but having a completely different technique in the benchmark is worthwhile I think.

> I removed it because it was very slow when querying, significantly affecting the total test time. Also, this method falls into a different category - compressed memory indexes for billion scale datasets, so I thought it was not much of interest anyway.

If I remember correctly, we actually used the raw vectors in their implementation, not the quantizied ones (otherwise accuracy was pretty low). So, it was just one example for the “inverted file” index, without compression. However, I think we actually had many useless parameter settings in the default `algos.yaml`. Maybe that’s where the slow query performance came from? For “good” parameter choices, faiss-ivf was around the same speed as annoy, so I would vote on getting it back. However, let’s wait for the next iteration. :-)

searchivarius:
@maumueller and @erikbern 
1. There has been a good set of data sets in this benchmark: https://github.com/DBWangGroupUNSW/nns_benchmark
2. We should also be running a sparse benchmark: a common belief that you can't use Annoy or similar approaches for the task, but you can. Use a hash trick to project the vectors into a sufficiently large dense one (if you have say 100 non-zero dim on average, something like 500-1000 should work), then use a regular cosine. I am really curious how far you can get with this simple dim. reduction approach.

erikbern:
> There has been a good set of data sets in this benchmark: https://github.com/DBWangGroupUNSW/nns_benchmark

These are great – might be worth picking a few of these!

searchivarius:
@erikbern but only sufficiently large one, please, not MNIST!

PR #84, opened 2018-06-26
Support for Batch mode

maumueller:
This PR improves the rudimentary support for batch mode that is already present in ann-benchmarks. Batch mode is enabled by providing `--batch` to `run.py`. Runs in non-batch/batch mode are treated completely independently by the the plotting scripts. That means `--batch` has to be provided to `plot.py` to plot results for batched runs; the website generated by`create_website.py` has different sections for batched/non-batched queries, where all batched runs have to suffix "-batch".

The approach is pretty straight-forward: There is a method `batch_query(X)` that gets the whole query set at once. The default implementation is to query each individual point with the standard query procedure, but all of faiss' and nmslib's implementations support a dedicated batch method that is used in this case. We can also run faiss' GPU implementation in this way.  

I fixed a few bugs on the way, see the individual commits. 

## Results for `random-xs-20-euclidean`:

Admittedly, not the most interesting dataset :-)

### Single queries

random-xs-20-euclidean

### Batched queries
random-xs-20-euclidean-batch

Of course, comparing a GPU implementation against CPU implementations is a bit odd here, but it clearly shows where the current bar for fast ANN algorithms is.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ before_install:
 
 script:
   - python run.py --docker-tag ann-benchmarks-${LIBRARY} --max-n-algorithms 5 --dataset random-xs-20-angular --run-disabled
+  - python run.py --docker-tag ann-benchmarks-${LIBRARY} --max-n-algorithms 5 --dataset random-xs-20-angular --run-disabled --batch
   - python plot.py --dataset random-xs-20-angular --output plot.png
+  - python plot.py --dataset random-xs-20-angular --output plot-batch.png --batch
   - python -m unittest test/test-metrics.py
   - python create_website.py --outputdir . --scatter --latex
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Running
 =======
 
 1. Run `python run.py` (this can take an extremely long time, potentially days)
-2. Run `python plot.py` to plot results.
+2. Run `python plot.py` or `python create_website.py` to plot results.
 
 You can customize the algorithms and datasets if you want to:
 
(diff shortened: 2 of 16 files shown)


erikbern:
wow, those are some impressive improvements! was that also run with gpu enabled?

as a side note i think it would be nice to default to logarithmic scale for the y axis, but not a big deal

install/Dockerfile.faiss
```diff
@@ -1,9 +1,9 @@
 FROM ann-benchmarks
 
-RUN apt-get update && apt-get install -y libopenblas-base libopenblas-dev libpython-dev python-numpy python-pip
+RUN apt-get update && apt-get install -y libopenblas-base libopenblas-dev libpython-dev python-numpy python-pip python-dev
 RUN git clone https://github.com/facebookresearch/faiss lib-faiss
-RUN cd lib-faiss && cp example_makefiles/makefile.inc.Linux makefile.inc && make -j4 py BLASLDFLAGS=/usr/lib/x86_64-linux-gnu/libopenblas.so.0
-ENV PYTHONPATH lib-faiss
+RUN export PYTHON=python && cd lib-faiss && cp example_makefiles/makefile.inc.Linux makefile.inc && ./configure && make -j4 BLASLDFLAGS=/usr/lib/x86_64-linux-gnu/libopenblas.so.0  && make py && make install
+ENV PYTHONPATH lib-faiss/python
```
erikbern: thanks for fixing this, have been meaning to!

maumueller:
> wow, those are some impressive improvements! was that also run with gpu enabled?

Not sure what you mean: `faiss-gpu` is run with a gpu enabled, yes.

install/Dockerfile.faiss
maumueller: Sure. It seems that they also support Python3 by now, but it needs some manual changes to the example makefile. Probably some `sed` magic would do the trick, but since it's running in docker it's not big deal anyways.

erikbern:
> Not sure what you mean: faiss-gpu is run with a gpu enabled, yes.

i meant – did you run it on a machine with a gpu?

maumueller:
>> Not sure what you mean: faiss-gpu is run with a gpu enabled, yes.

> i meant – did you run it on a machine with a gpu?

Yes, p2.xlarge.

maumueller:
No worries, I will resolve the merge problems tomorrow.

maumueller:
Hi again, I've decided to wait until #93 is done and merged. Since I made some changes to the plotting scripts there, merging it now only results in new conflicts. Same for #85.

erikbern:
want to rebase this one and i can merge it?

maumueller:
Done. 

Travis fails because of a problem with PyNNDescent, any ideas @lmcinnes ?

erikbern:
Thanks @maumueller !

pynndescent is failing on this right now: `Inconsistency detected by ld.so: dl-version.c: 205: _dl_check_map_versions: Assertion `needed != NULL' failed!`

lmcinnes:
I believe that is an issue introduced by the latest version of numba. I can push a workaround fix until numba resolves the main issue.

PR #85, opened 2018-06-26
Bitvectors and Hamming distance

maumueller:
This PR adds (back) support for bitvectors under Hamming distance. Note that it requires to store the `point_type` within the hdf5 file, so vectors have to be regenerated, @erikbern.

I have added a binary version of the SIFT dataset to it, generated using Spherical hashing. Unfortunately, I don't know of a nice Python implementation of Spherical hashing which would allow producing the binary embedding from SIFT directly from the source (and for other of our datasets as well).

I've tried to add back support for set similarity under Jaccard distance.  Here, each data point is a variable-length set of integers. This turns out to be difficult with the current infrastructure: First, hdf5 doesn't support vectors of `frozenset`s in a straight-forward way. Next, the code becomes really ugly because of the current print outs of the current state including the dimensionality of the vectors, such as:
```python3
 def write_output(train, test, fn, distance, point_type='float', count=100):
       from ann_benchmarks.algorithms.bruteforce import BruteForceBLAS
       n = 0
       f = h5py.File(fn, 'w')
       f.attrs['distance'] = distance
       f.attrs['point_type'] = point_type
       print('train size: %9d * %4d' % train.shape)
       print('test size:  %9d * %4d' % test.shape)
```
For sets, there is no fixed dimensionality and so we would have to distinguish each of the printouts by the type of the dataset at hand. I'm a bit unsure how a nice solution would look like.

PS: This is independent of #84 and both can be merged into each other.

### DIFF
--- a/ann_benchmarks/datasets.py
+++ b/ann_benchmarks/datasets.py
@@ -236,7 +236,7 @@ def sift_hamming(out_fn, fn):
         lines = f.readlines()
         X = numpy.zeros((len(lines), 256), dtype=numpy.bool)
         for i, line in enumerate(lines):
-            X[i] = numpy.array([int(x) > 0 for x in line.strip()], dtype=numpy.bool)
+            X[i] = numpy.array([int(x) > 0 for x in line.decode().strip()], dtype=numpy.bool)
         X_train, X_test = train_test_split(X, test_size = 1000)
         write_output(X_train, X_test, out_fn, 'hamming', 'bit')
 
--- a/ann_benchmarks/datasets.py
+++ b/ann_benchmarks/datasets.py
@@ -222,9 +222,23 @@ def word2bits(out_fn, path, fn):
         for i in range(n_words):
             X[i] = numpy.array([float(z) > 0 for z in next(f).strip().split()[1:]], dtype=numpy.bool)
 
-        X_train, X_test = train_test_split(X, test_size=50)
+        X_train, X_test = train_test_split(X, test_size=1000)
         write_output(X_train, X_test, out_fn, 'hamming', 'bit')
 
+def sift_hamming(out_fn, fn):
+    import tarfile
+    local_fn = fn + '.tar.gz'
+    url = 'http://sss.projects.itu.dk/ann-benchmarks/datasets/%s.tar.gz' % fn
+    download(url, local_fn)
+    print('parsing vectors in %s...' % local_fn)
+    with tarfile.open(local_fn, 'r:gz') as t:
+        f = t.extractfile(fn)
+        lines = f.readlines()
+        X = numpy.zeros((len(lines), 256), dtype=numpy.bool)
+        for i, line in enumerate(lines):
+            X[i] = numpy.array([int(x) > 0 for x in line.strip()], dtype=numpy.bool)
+        X_train, X_test = train_test_split(X, test_size = 1000)
+        write_output(X_train, X_test, out_fn, 'hamming', 'bit')
 
 def lastfm(out_fn, n_dimensions, test_size=50000):
     # This tests out ANN methods for retrieval on simple matrix factorization based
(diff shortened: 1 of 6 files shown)


erikbern:
cool that this was so little code

i think you can also just use the dtype of the dataset in hdf5 rather than writing it as an attr. pyh5 supports boolean types.

maumueller:
> cool that this was so little code

Luckily, most other pieces where left in the code ;-)

> i think you can also just use the dtype of the dataset in hdf5 rather than writing it as an attr. pyh5 supports boolean types.

That's what I started out with. I wrote it as an attribute when I tinkered around with Jaccard similarity, because the custom dtype that I'd used there made the code look really ugly.

I can roll that back, but I think writing it into the file is the more generic approach.

erikbern:
> I can roll that back, but I think writing it into the file is the more generic approach.

seems fine. you can also default to `float` if there's no attribute set (that way i don't have to regenerate all the datasets, but maybe i'm a bit too lazy/selfish)

maumueller:
That seems fair enough, done.

plot.py
```diff
@@ -94,7 +94,9 @@ def create_plot(all_data, raw, x_log, y_log, xn, yn, fn_out, linestyles):
 
     dataset = get_dataset(args.dataset)
     dimension = len(dataset['train'][0]) # TODO(erikbern): ugly
-    point_type = 'float' # TODO(erikbern): should look at the type of X_train
+    point_type = 'float'
+    if 'point_type' in dataset.attrs:
+        point_type = dataset.attrs['point_type']
```
erikbern: i think you can just do `dataset.attrs.get('point_type', 'float')`

maumueller: 👍 Done

ann_benchmarks/datasets.py
```diff
@@ -219,11 +220,25 @@ def word2bits(out_fn, path, fn):
...
 
-        X_train, X_test = train_test_split(X)
-        write_output(X_train, X_test, out_fn, 'euclidean')  # TODO: use hamming
+        X_train, X_test = train_test_split(X, test_size=1000)
+        write_output(X_train, X_test, out_fn, 'hamming', 'bit')
 
+def sift_hamming(out_fn, fn):
+    import tarfile
+    local_fn = fn + '.tar.gz'
+    url = 'http://sss.projects.itu.dk/ann-benchmarks/datasets/%s.tar.gz' % fn
```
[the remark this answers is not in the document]
maumueller: See above: 

> I have added a binary version of the SIFT dataset to it, generated using Spherical hashing. Unfortunately, I don't know of a nice Python implementation of Spherical hashing which would allow producing the binary embedding from SIFT directly from the source (and for other of our datasets as well).

It's [this method](http://www.ee.columbia.edu/ln/dvmm/pubs/files/Spherical_Hashing.pdf) using their implementation from https://github.com/jp-sglab/Spherical_Hashing.

erikbern:
Looks nice! Will take a look later tonight

maumueller:
Rebased this one to current master as well. Tiny additional change in https://github.com/erikbern/ann-benchmarks/pull/85/commits/3d4290f53f27a2c5c70c30a2d8743991b6bcb6a4 (got really frustrated that there is no fallback when a dataset hasn't been uploaded to s3). Can leave that out if you want.

erikbern:
Why don't you just upload it to S3? Which dataset is not in S3 right now?

I'd rather keep all datasets in S3 and not fall back on creating them on the fly. I guess it's fine if the fallback is to use any local file though... don't have super strong feelings though, so I'm fine merging this

maumueller:
Ok, uploaded the lastfm and hamming datasets to S3. All datasets that we actually use in the tests should be on S3 now. I'd like to see the failback stay there! Ready to be merged from my side.

erikbern:
> Ok, uploaded the lastfm and hamming datasets to S3. All datasets that we actually use in the tests should be on S3 now. I'd like to see the failback stay there! Ready to be merged from my side.

Thanks for uploading! Can you also link to them from README.md?

I would have a mild preference for not using fallbacks (feels like it increases the risk of unintended consequences) but not a big deal to me really. Let's merge 🚢

erikbern:
Thanks for the contribution @maumueller 

Will include these in next benchmarks!

PR #86, opened 2018-06-29
update links to use ann-benchmarks.com instead of vectors.erikbern.com

erikbern:


erikbern:
should have merged this a long time ago!

PR #87, opened 2018-07-02
Choose specific tag for avoiding error about lib-faiss

Writtic:
- Choose specific tag for avoiding error from master branch of faiss
- Fix typo of pip install -rrequirements.txt


erikbern:
thanks @Writtic !

This this is also fixed in #84 btw, but might as well merge this one separately

PR #88, opened 2018-07-10
Add multiprocessing for building multiple docker images simultaneously

Writtic:
- Add `--proc` option for setting multiprocessing pool to build docker
images faster
- Add `--algorithm` option seeming like `LIBRARY` environment
- Remain `LIBRARY` env option but `--algorithm` option has higher priority

### DIFF
--- a/install.py
+++ b/install.py
@@ -41,9 +41,12 @@ if __name__ == "__main__":
         build(os.getenv('LIBRARY'))
     else:
         print('Building algorithm images... with (%d) processes' % args.proc)
-        pool = Pool(processes=args.proc)
         dockerfiles = []
         for fn in os.listdir('install'):
             if fn.startswith('Dockerfile.'):
                 dockerfiles.append(fn.split('.')[-1])
+
+        pool = Pool(processes=args.proc)
         pool.map(build, dockerfiles)
+        pool.close()
+        pool.join()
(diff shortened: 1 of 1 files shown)


maumueller:
Nice @Writtic! I had such a solution hiding in a local branch as well.

Note that there is a `positive_int` function present in https://github.com/erikbern/ann-benchmarks/blob/b8e0e0e8c7dd078ec1710bdfde097e67de80f5cd/ann_benchmarks/main.py#L16-L24

Why not just use that one?

erikbern:
This makes sense. Just to double check did you see a substantial speed up running this locally? Asking since I think docker should generally take advantage of multiple cores already when the underlying program supports it

Writtic:
I just checked and compared multiprocessed docker building time except for base docker image building time. Then here is the result. (In the test result ANN Algorithm of My team was included)
```
CPU: Intel(R) Xeon(R) Silver 4110 CPU @ 2.10GHz x 8 cores (32 logical processors)
MEM: 252GB

python install.py --proc=32
--- 728.110238 sec ---

python install.py --proc=12
--- 752.668900 sec ---

python install.py --proc=4
--- 840.305218 sec ---

python install.py --proc=2
--- 958.391327 sec ---

python install.py (original)
--- 1730.246785 sec ---
```

According to the results, it seems to be worth. I suppose Docker filesystem could be bottleneck of building process but multiprocessing is not a concerns. Docker doesn't takes some advantage of multiple cores unless build script directly use it. (Maybe something like make -j4  can use multiple cores but almost building sequence doesn't use those)

erikbern:
Thanks @writtic !

PR #91, opened 2018-08-05
add lastfm dataset

benfred:
This adds a dataset based off the lastfm360K dataset, using factors generated by a simple MF based recommendation model.

This lets us replicate the experiments done in this post: http://www.benfrederickson.com/approximate-nearest-neighbours-for-recommender-systems/ , with the main benefit that this lets us test out inner product lookup of MF based recommender models on top of angular ANN indices.

The generated dataset is here https://github.com/benfred/recommender_data/releases/download/v1.0/lastfm-64-dot.hdf5


erikbern:
Sorry for the late realization but I never linked to this from the README. adding that now as a part of #166

PR #92, opened 2018-08-21
add ONNG

masajiro:
We added our ONNG(NGT) which is based on a different method from our PANNG(NGT) to ann benchmarks.
We also reduced the total processing time for PANNG by eliminating index construction for each query parameter.

Thank you in advance.


erikbern:
Nice!

I will include it next time I run the benchmarks (probably in a few months)

Issue #73, opened 2018-05-07
Zero vectors in nytimes-256-angular


yurymalkov:
I found that nytimes-256-angular and nytimes-16-angular datasets have hundreds of zero-valued vectors.
The problem is that the cosine distance to such vectors is undefined, making the search an ill-defined task.
Probably, such vectors should be removed. Another option is using unnormalized inner product distance instead of cosine.

erikbern:
I don't have super strong feelings about this, but think this is arguably a "feature" not a bug. Vectors that are zero should simply be ignored and never returned.

Annoy actually didn't do this and had very bad performance for nytimes-256-angular until I fixed it.

yurymalkov:
If it is a feature, it's probably a not very good design choice.
Cosine similarity for null vectors is fundamentally undefined, some discussion can be found here https://stackoverflow.com/questions/26699659/cosine-similarity-when-one-of-vectors-is-all-zeros . 
So one can expect general-purpose search algorithms to rightfully show different behavior on such data.

yurymalkov:
Ok. I am busy now with another pull request to this repo (adding structure reusing to some algorithms). Maybe I'll do it later.

erikbern:
ok awesome!

PR #93, opened 2018-08-27
Added ONNG results.

maumueller:
After #92 I was interested to see how ONNG actually performs, and it performs quite well! (Much better than their other variant.) Find the updated result plot files in the PR and on http://ann-benchmarks.com. 

One problem with the plot script in this situation: The change done to PANNG  in `algos.yaml` by https://github.com/erikbern/ann-benchmarks/commit/0b50eb2f149bd0dee343d8c8274b2d3b8d7827b8 made the PANNG result files inaccessible to `plot.py`, since their names changed. I dislike this tight connection between the algos.yaml file and the result files. The plotting done in `create_website.py` just enumerates all result files and plots them, which is much more robust. Maybe we should update `plot.py` to use the same approach?


erikbern:
> The plotting done in create_website.py just enumerates all result files and plots them, which is much more robust. Maybe we should update plot.py to use the same approach?

Seems fine to me, if we can factor that out into a helper function that can be used by both use cases

erikbern:
great results!

maumueller:
>> The plotting done in create_website.py just enumerates all result files and plots them, which is much more robust. Maybe we should update plot.py to use the same approach?

> Seems fine to me, if we can factor that out into a helper function that can be used by both use cases

I'll give it a try, but probably not before late in this week. 

> great results!

I agree. It's a pity that preprocessing in our single-thread setup is so slow for their approach. Would be great to see how it performs on `gist` and `nytimes`. @masajiro, maybe you could provide preprocessing parameters for these datasets? It seems that it ought to be roughly 2 times faster to finish these datasets in the 5h time limit.

masajiro:
@maumueller 
Thank you for plotting ONNG.
> maybe you could provide preprocessing parameters for these datasets?

I will investigate better parameters to shorten the construction time. However, it is a very difficult balancing act, because, in most cases, shortening the construction time tends to reduce the search performances.

masajiro:
@maumueller 
> maybe you could provide preprocessing parameters for these datasets?

At this moment, the easiest way to reduce construction time is decreasing the parameter **edge** which is the first parameter in [this line](https://github.com/erikbern/ann-benchmarks/blob/master/algos.yaml#L262). Could you try to change it from 100 to 80 or 60?

BTW, for your plotting, there is something strange. The performances of ONNG for fashion-mnist-784 are worse than our benchmarks comparing to hnsw. Did you have any problem during the benchmarks, or do you have any clue about that around your environment.

maumueller:
@masajiro 

> > maybe you could provide preprocessing parameters for these datasets?

> At this moment, the easiest way to reduce construction time is decreasing the parameter edge which is the first parameter in this line. Could you try to change it from 100 to 80 or 60?

Experiments are running right now. I'll report back.

> BTW, for your plotting, there is something strange. The performances of ONNG for fashion-mnist-784 are worse than our benchmarks comparing to hnsw. Did you have any problem during the benchmarks, or do you have any clue about that around your environment.

Rerunning these experiments as well.

masajiro:
@maumueller 
I created the pull request #98 to resolve too long build time. If possible, could you please try to use it?

maumueller:
@masajiro The results have been on my ec2 instance, I just forgot about collecting them and plotting them. I will look at that tomorrow and report back. Then we can give it another try if necessary. Thanks for #98.

masajiro:
@maumueller The previous ONNG cannot build expected indexes for the NYTimes due to the issue #73. I would appreciate if you could try again at least for NYTimes.

maumueller:
@masajiro 

>>> maybe you could provide preprocessing parameters for these datasets?

>> At this moment, the easiest way to reduce construction time is decreasing the parameter edge which is the first parameter in this line. Could you try to change it from 100 to 80 or 60?

> Experiments are running right now. I'll report back.

Almost all of the experiments timed-out with 80. Running them again with #98 included. Should it work with `edge` set to 100? 

>> BTW, for your plotting, there is something strange. The performances of ONNG for fashion-mnist-784 are worse than our benchmarks comparing to hnsw. Did you have any problem during the benchmarks, or do you have any clue about that around your environment.

> Rerunning these experiments as well.

Looks like this:

fashion-mnist-784-euclidean

Is that similar to your local experiments?

maumueller:
@erikbern I could also include the results for `shidx` in this PR. Is that fine? (Doesn't seem competitive, as above.)

maumueller:
Updated plots for ONNG. Everything went through smoothly, @masajiro . Great results. 

Ready to merge from my side.

masajiro:
@maumueller Thank you for dealing with my requests!

> > > > maybe you could provide preprocessing parameters for these datasets?
> 
> > > At this moment, the easiest way to reduce construction time is decreasing the parameter edge which is the first parameter in this line. Could you try to change it from 100 to 80 or 60?
> 
> > Experiments are running right now. I'll report back.
> 
> Almost all of the experiments timed-out with 80. Running them again with #98 included. Should it work with `edge` set to 100?

100 is fine for the `edge`, because the latest ONNG automatically reduces the edge parameter, if necessary.

> > > BTW, for your plotting, there is something strange. The performances of ONNG for fashion-mnist-784 are worse than our benchmarks comparing to hnsw. Did you have any problem during the benchmarks, or do you have any clue about that around your environment.
> 
> > Rerunning these experiments as well.
> 
> Looks like this:
> 
> 
> Is that similar to your local experiments?

Yes, this result is similar to my local result.

erikbern:
can i go ahead and merge?

maumueller:
I suppose that is a question to @masajiro?

erikbern:
well both but let's just do it :)

masajiro:
> I suppose that is a question to @masajiro?

@maumueller @erikbern 
My answer is too late?
The ONNG result is definitely lower than my result for the nytimes.I’m not sure the exact reason, but there is a possibility that fewer plots might tend to cause unstable results due to not being cached.

If I increase the plots, could you try nytimes again?

maumueller:
@masajiro -- What do you mean by "plots"? I can try again if you tell me what to test for.

masajiro:
@maumueller 
I would like to increase the number of plotted points in the figure. 
Could you replace the line below 
https://github.com/erikbern/ann-benchmarks/blob/master/algos.yaml#L280
with the line below?
```
query-args : [[0.6, 0.8, 0.9, 0.95, 0.98, 1.0, 1.02, 1.04, 1.06, 1.07, 1.08, 1.09, 1.1, 1.12, 1.14, 1.16, 1.18, 1.2, 1.25]]
```
Or I can create a pull request for this update.

maumueller:
Situation for "medium recall" did not improve by another run with the given parameter settings: 
nytimes-256-angular

masajiro:
@maumueller 
Thank you for your cooperation to resolve the problem!

The previous points remain in the graph, don't they?
Could you remove all of the existing results and give it a run by using this parameter below?

```
query-args : [[0.6, 0.8, 0.9, 0.95, 0.98, 0.99, 0.995, 1.0, 1.01, 1.02, 1.03, 1.04, 1.05, 1.06, 1.07, 1.08, 1.09, 1.1, 1.12, 1.14, 1.16, 1.18, 1.2, 1.25]]
```
Thank you in advance.

masajiro:
@maumueller 
I could not find low recalls for 100 NN search in my experiments. But I found low recalls for 10 NN search due to invalid zero vectors in the query set as well. I will deal with the zero vectors of the query set soon.

PR #95, opened 2018-08-31
build-arg for install.py. Support for python 2.x in invoking Thread…

eruffaldi:
… in runner.py


erikbern:
I'm skeptical of supporting Python 2 but I guess this PR is small enough that it probably doesn't hurt :)

erikbern:
thanks!

PR #100, opened 2018-10-26
Subprocess support

ale-f:
Here's the new and tidied-up version of the last(?) thing we left out of #37; the text-based protocol for running experiments in subprocesses! There's now much more documentation for the protocol, as well as extensions for it that add support for batch queries and query argument groups.

Here are some numbers. I wrote a little Python wrapper that runs `BruteForce` and `BruteForceBLAS` using the protocol, and in normal mode the protocol's about 75% of the speed of the normal Python version:

normal

Something really weird happens in batch mode, though, where it runs at _105%_ of the speed:

batch

I suppose it's just because the naïve implementation of batch mode has to build a Python array internally, but still -- in the right mode, there's more or less no overhead, which is (unexpectedly) nice.

@maumueller, you had some examples of algorithms you wanted to include in this way, didn't you?


erikbern:
thanks – will review this weekend.

looks like it doesn't affect the core of ann-benchmarks much so should be fine to merge

are there any tests? can you add any of the libraries that are using this protocol?

ale-f:
Oh, right, yeah; there was _supposed_ to be another algorithm definitions file in here (it's the one I used to make those plots), but I left that out by mistake. It's in there now, and it's in the Travis manifest as well.

@maumueller does indeed have some libraries to include; there's an old pull request in my repository. It needs a bit of updating before it can be included here, though!

protocol/bf-runner.py
```diff
@@ -0,0 +1,196 @@
+from __future__ import print_function
```
erikbern: can you reformat this to have consistent indentation (4 spaces)?

thanks

erikbern:
nice, will take a look at this again later today and hopefully merge

.travis.yml
```diff
@@ -35,3 +35,6 @@ script:
   - python plot.py --dataset random-xs-20-angular --output plot-batch.png --batch
   - python -m unittest test/test-metrics.py
   - python create_website.py --outputdir . --scatter --latex
+  - python run.py --definitions algosP.yaml --dataset random-xs-20-angular --local --runs 20
```
erikbern: I'm not sure why this has to run as a part of every library.

Why not just make it its own separate Travis run?

But why don't I merge this and just fix it myself since this PR has been sitting open for a while

algosP.yaml
```diff
@@ -0,0 +1,67 @@
...
+      module: ann_benchmarks.algorithms.bruteforce
+      constructor: BruteForceBLAS
+      base-args: ["@metric"]
+      run-groups:
+        empty:
+          args: []
+  angular:
+    pp-bruteforce-lo:
+      module: ann_benchmarks.algorithms.subprocess
+      docker-tag: ann-benchmarks-subprocess
```
erikbern: `install/Dockerfile.subprocess` is missing though. Did you forget to include that file?

ale-f: Good point; we never actually *wrote* that file! `ann_benchmarks/algorithms/definitions.py` now requires that every algorithm have a `docker-tag` property, so I just put a value in there to identify this group.

ale-f:
I should say, incidentally, that @maumueller revised one of those pull requests: https://github.com/ale-f/ann-benchmarks/pull/5. Do you want me to include that in here? (It's kind of nice how small the change is to add a new implementation...)

erikbern:
that would be great – why don't you include  https://github.com/ale-f/ann-benchmarks/pull/5 in this PR and then you can remove the extra steps in `.travis.yml` and make it a separate test environment instead (`LIBRARY=mih` or something similar)

ale-f:
Bah -- MIH works in Hamming space, and there aren't any small bit-vector datasets we could reasonably use to test it. @maumueller and I talked a bit about adding code to `create_dataset.py` to create a dataset of random bit vectors; we might try to get that done this week.

erikbern:
cool, lmk when this is ready and i'll re-review

had a baby earlier this week so will probably be very slow to get back right now 👶

ale-f:
Oh! Congratulations :-)

I think that's everything in reasonable shape here (and, at long last, Travis agrees), so take a look at your convenience.

erikbern:
Nice!

Issue #83, opened 2018-06-23
Regarding an interesting project


ghost:
@erikbern do you have any thoughts or observations about the `kepler-mapper`?
https://github.com/MLWave/kepler-mapper
https://mlwave.github.io/kepler-mapper/examples.html

PR #102, opened 2018-12-05
Add autotuning and update proper parameters for MRPT

vioshyvo:
We updated the benchmarking code of MRPT to match a new version of MRPT which has automatic tuning of hyperparameters. This speeds up benchmarking significantly, since there is no need to perform a grid search over 3 hyperparameters anymore.  

The parameters currently used for MRPT are really suboptimal: maximum number of trees 100 and maximum depth is 8, but for larger datasets, such as SIFT and GIST, the optimal number of trees is usually on the range 200-1000, and the optimal depth on the range 8-12. This PR also fixes this.

Now MRPT also works for angular distance (see comments on #83). The problem was in our python bindings.

### DIFF
--- a/algos.yaml
+++ b/algos.yaml
@@ -284,10 +284,10 @@ float:
       constructor: MRPT
       base-args: ["@metric"]
       run-groups:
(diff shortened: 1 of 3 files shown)


erikbern:
Nice!

PR #103, opened 2018-12-05
Remove duplicates when creating nytimes data set

### LINKED ISSUE #73: Zero vectors in nytimes-256-angular

vioshyvo:
There are over 40000 duplicates in the original nytimes data set. This makes it not a very good benchmark data set, since on the one hand these would probably be removed in normal preprocessing, and on the other hand if there are more than k duplicates for a query point, the indices of k nearest neighbours are not uniquely determined anymore.

So I propose that we just remove the duplicates when creating the data set.

This would also fix #73.

### DIFF
--- a/ann_benchmarks/datasets.py
+++ b/ann_benchmarks/datasets.py
@@ -197,6 +197,7 @@ def transform_bag_of_words(filename, n_dimensions, out_fn):
         B = TfidfTransformer().fit_transform(A)
         print("reducing dimensionality...")
         C = random_projection.GaussianRandomProjection(n_components = n_dimensions).fit_transform(B)
+        C = numpy.unique(C, axis = 0) # remove duplicates
         X_train, X_test = train_test_split(C)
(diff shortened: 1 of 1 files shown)


erikbern:
Probably makes sense, thanks @vioshyvo

lmcinnes:
The duplicates are actually not a bad stress test for algorithms; such things certainly do happen in practice. Notably because of the way accuracy is measured based on distance rather than indices the duplicates are interchangeable from an accuracy standpoint, so as long as you return the right number of duplicates all is well. The zero vectors (and there are many even in the test/query set) are more troubling, particularly as they are duplicated. It is effectively impossible to get them correct, so there is a maximum possible accuracy for nytimes.

PR #105, opened 2019-02-13
update onng parameter

masajiro:
I adjusted a value of the ONNG’s parameter for the latest ONNG version (v1.6.2).
Thank you in advance.


erikbern:
not sure why flann test is failing

thanks @masajiro

PR #107, opened 2019-03-06
Small changes to metrics computation + additional metrics

maumueller:
This PR makes the following changes:

- Added mechanism to allow an implementation to report on additional characteristics. This is showcast by having IVF and HNSW (from faiss) report the number of distance computations done to answer the set of queries (which is a general benchmark of the quality more independent than running times).
- Small refactorings in metric computation, report not only on avg but also standard deviation. (To be later added as error bars in plots.)
- Store computed metrics inside the result file to avoid recomputation
- Improved plotting time by a factor ~4 by better caching of hdf5 access and small changes to the metric computation

The general goal of these changes is to make it easier to work with our result files, e.g. for visualization of results beyond those provided at the moment.  

We use it to create plots like this, which provide more inside into what the distribution of individual recall values looks compared to the average (the single dot):
image

This gives more details into the inner workings of different algorithmic ideas. (To be discussed:  should such plottings scripts be added to ann-benchmarks as well?)

### DIFF
--- a/ann_benchmarks/plotting/metrics.py
+++ b/ann_benchmarks/plotting/metrics.py
@@ -1,31 +1,64 @@
 from __future__ import absolute_import
+import numpy as np
 
-def knn(dataset_distances, run_distances, count, epsilon=1e-10):
+def get_recall_values(dataset_distances, run_distances, count, threshold):
     total = len(run_distances) * count
-    actual = 0
-    for true_distances, found_distances in zip(dataset_distances, run_distances):
-        within = [d for d in found_distances[:count] if d <= true_distances[count - 1] + epsilon]
-        actual += len(within)
-    return float(actual) / float(total)
+    recalls = np.zeros(len(run_distances))
+    for i in range(len(run_distances)):
+        t = threshold(dataset_distances[i])
+        actual = 0
+        for j in range(count):
+            if run_distances[i][j] <= t:
+                actual += 1
+            else:
+                break
+        recalls[i] = actual
+    return np.mean(recalls) / float(count), np.std(recalls) / float(count), recalls
 
-def epsilon(dataset_distances, run_distances, count, epsilon=0.01):
-    total = len(run_distances) * count
-    actual = 0
-    for true_distances, found_distances in zip(dataset_distances, run_distances):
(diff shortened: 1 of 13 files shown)


ann_benchmarks/plotting/metrics.py
```diff
@@ -1,31 +1,70 @@
...
-    actual = 0
-    for true_distances, found_distances in zip(dataset_distances, run_distances):
-        within = [d for d in found_distances[:count] if d <= true_distances[count - 1] * (1 + epsilon)]
-        actual += len(within)
-    return float(actual) / float(total)
+    recalls = np.zeros(len(run_distances))
+    for i in range(len(run_distances)):
+        t = threshold(dataset_distances[i], count, epsilon)
+        actual = 0
+        for j in range(min(count, len(run_distances[i]))):
```
erikbern: `for j, d in enumerate(run_distances[i][:count]):` :)

erikbern:
i think this change looks good – i don't find the plots super easy to understand, but the code seems fine!

erikbern:
looks like build is passing except for flann (will solve separately)

ann_benchmarks/plotting/metrics.py
maumueller: Thanks! Much better. Changed it in cdda15b. (Also removed the break statement because it assumes that algorithms sort their answers to queries.)

maumueller:
> looks like build is passing except for flann (will solve separately)

That seems to be related to https://github.com/mariusmuja/flann/pull/399. I've added the libraries added their as explicit requirements in ce74df9, but their seems to be another issue with the Python wrapper.

maumueller:
> i think this change looks good – i don't find the plots super easy to understand, but the code seems fine!

A caption is definitely necessary. ;-) We made quite a lot of new plots for a recent paper. I don't think that all of such scripts should go into the base repo. What do you think about a repo just for additional plotting functionality? (`ann-benchmarks-plotting`? Would be easiest if everything would go under an ann-benchmarks organization.)

maumueller:
Flann is still not building properly, opened up https://github.com/mariusmuja/flann/issues/406.

erikbern:
thanks! i might disable flann for now

PR #114, opened 2019-04-12
Fix Pareto frontier generation.

MIN605472:
If we have a set of points that have the same Y value, those that are more valuable on X should be first.

The error is most noticeable right now in the Index size-Recall tradeoff graphs where sometimes there are line segments parallel to the recall axis (different recall values having the same index size).


erikbern:
cool – i trust that this works :)

PR #121, opened 2019-06-24
Update results

masajiro:
I built all of the containers on 10th Jun and started them. These are the results. After the benchmark, although I ran sptag as well, I could not get its results because of the timeout.


maumueller:
Thanks @masajiro for rerunning the benchmark!

In general, I am not happy with just sharing the png's. I would rather hope that you could share the raw experimental data with us as well so that we can reproduce the plots ourselves and update the website at http://ann-benchmarks.com/. (Only now I've noticed that the raw result files are not available there as well...). 

@erikbern What do you think?

erikbern:
agree with that @maumueller

masajiro:
@maumueller I added the raw results as well.

maumueller:
Thanks @masajiro! I'm honestly surprised that Github allowed you to do that! 

I don't think it's a good idea to put these files into version control. Everybody has to check out a 1 GB repo just to run ann-benchmarks if this ever gets merged. However, I checked out your repo to get the raw result files. I propose that I'll put them somewhere else and update the website + the png plots later today.

masajiro:
@maumueller In terms of version control, I probably should not have added the raw results.

BTW, even though you use c5.4xlarge, since the instance where you will run is different from the instance where I produced the raw results, I think that you should run all of the methods again to compare. The difference of the instances causes slight but not negligible different results to compare.

maumueller:
I see, @masajiro! For reproducibility reasons, the official results should be run on the same instance. I noticed some (absolute) differences when I re-run some of the results (Note that http://ann-benchmarks.com/ contains a few more datasets/# of NN). 

I would suggest that we close this PR without merging and re-run all experiments at some point. You can of course publish the results of your run in the ONNG repo or at other places. 

@erikbern It's currently really a pain to have single threaded index building for all the graph-based approaches. It feels so wasteful to pay for all the cores, when we actually only need the RAM resources. Any way we allow index building to be multi-threaded again? (It seems straight-forward to set querying to be single-threaded for the libraries.)

maumueller:
Hi again!

After some e-mail discussions with @masajiro, I now agree with the PR (if the raw results are excluded again). I initially thought he didn't run it on ec2, which he did. That's on me.

@erikbern: @masajiro provided the raw results and is currently running the 100-NN tests for ann-benchmarks.com. I checked that the raw results provided by him produce the plots in this PR. I think this can be merged and I'll update the website as soon as the 100-NN results are available (which will take around 2 more weeks). What do you think?

erikbern:
sounds good, just force-push a new changeset which doesn't include the raw results just so we avoid having a multi-gigabyte commit in the repo

masajiro:
@maumueller Thank you for your comments.
@erikbern I deleted the raw results with force-pushed.

erikbern:
cool, i'm ok merging this if @maumueller is good w it

maumueller:
> cool, i'm ok merging this if @maumueller is good w it

:+1:

PR #119, opened 2019-06-11
add SPTAG algorithm

SueSu-Wish:
SPTAG: https://github.com/microsoft/SPTAG

PR #124, opened 2019-07-16
Updated SPTAG docker file

maumueller:
As discussed in #119, this PR makes SPTAG use the libboost1.67-dev-all package instead of building boost itself.

libboost-dev-all comes with a Python 2 dependency, which is the preferred Python version in SPTAG's build system. Thus I changed the entry point to run with Python 2.

### DIFF
--- a/algos.yaml
+++ b/algos.yaml
@@ -1,5 +1,14 @@
 float:
   any:
+    sptag:
+      docker-tag: ann-benchmarks-sptag
+      module: ann_benchmarks.algorithms.sptag
+      constructor: Sptag
+      base-args: ["@metric"]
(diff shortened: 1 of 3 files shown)


erikbern:
thanks!

tekumara:
SPTAG builds with both Python 2 & Python 3 see https://github.com/microsoft/SPTAG/blob/053c04fce2b0f013aaef25ec52d8c50d3692af49/Wrappers/CMakeLists.txt

Python 2 is a little dated now, so SPTAG Dockerfile uses Python 3 with boost built from source.
see
https://github.com/microsoft/SPTAG/blob/master/Dockerfile

PR #126, opened 2019-07-16
Fix naming of Faiss' HNSW and SPTAG.

maumueller:
Adds a missing query parameter to the name in the result file. Currently we cannot say which query parameter gave a particular data point.


maumueller:
I noticed the same problem with SPTAG, so I fixed it there as well. (+ avoided a call `tobytes()` which is not present in https://github.com/microsoft/SPTAG/blob/master/docs/GettingStart.md.)

erikbern:
thanks

wonder what caused sptag to start failing on travis

PR #127, opened 2019-07-19
Reintroduce Set Similarity under Jaccard Distance

maumueller:
This PR reintroduces support for ANN for sets under Jaccard similarity. (It always bothered me that we lost support for a 1k star ANN implementation https://github.com/ekzhu/datasketch/.) Currently, there is only one standard data set and some random sets for testing.

I tried to keep the PR as minimally invasive as possible. This means that sets are stored as bit vector representation in hdf5 (because they are supposed to be 2d np matrices), but are then transformed to a sparse list-of-indices representation (needed by the algorithms).

Looking forward to hearing your thoughts!

### DIFF
--- a/ann_benchmarks/datasets.py
+++ b/ann_benchmarks/datasets.py
@@ -8,6 +8,8 @@ try:
 except ImportError:
     from urllib.request import urlretrieve  # Python 3
 
+from ann_benchmarks.distance import dataset_transform
+
 
 def download(src, dst):
     if not os.path.exists(dst):
@@ -55,11 +57,13 @@ def write_output(train, test, fn, distance, point_type='float', count=100):
     neighbors = f.create_dataset('neighbors', (len(test), count), dtype='i')
     distances = f.create_dataset('distances', (len(test), count), dtype='f')
     bf = BruteForceBLAS(distance, precision=train.dtype)
+    train = dataset_transform[distance](train)
+    test = dataset_transform[distance](test)
     bf.fit(train)
     queries = []
     for i, x in enumerate(test):
         if i % 1000 == 0:
-            print('%d/%d...' % (i, test.shape[0]))
+            print('%d/%d...' % (i, len(test)))
         res = list(bf.query_with_distances(x, count))
         res.sort(key=lambda t: t[-1])
         neighbors[i] = [j for j, _ in res]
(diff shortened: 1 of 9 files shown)


maumueller:
The current leader board is a bit empty, so it might not be worth adding it to the README :-)

kosarak-jaccard

erikbern:
nice! i think jaccard could probably be added to annoy in some hacky way.

ann_benchmarks/datasets.py
```diff
@@ -279,6 +283,50 @@ def sift_hamming(out_fn, fn):
...
+    for line in content:
+        if len(line.split()) >= min_elements:
+            for x in line.split():
+                X[i][ids[int(x)]] = 1
+            i += 1
+
+    X_train, X_test = train_test_split(numpy.array(X), test_size=500)
+    write_output(X_train, X_test, out_fn, 'jaccard', 'bit')
+
+def random_jaccard(out_fn, n=10000, size=50, universe=80):
```
erikbern: can you seed the random here just so it's reproducible?

ann_benchmarks/distance.py
```diff
@@ -1,20 +1,38 @@
...
-    intersect = len(a & b)
+         return 0
+    intersect = len(set(a) & set(b))
     return intersect / (float)(len(a) + len(b) - intersect)
 
+def transform_dense_to_sparse(X):
+    """Converts the n * m dataset into a sparse format
+    that only holds the non-zero entries (Jaccard)."""
+    # get list of indices of non-zero elements
+    indices = np.transpose(np.where(X))
```
erikbern: not super important but you could probably use `itertools.groupby` here:
```python
indices = np.transpose(np.where(X))
keys = []
for ijs in itertools.groupby(indices,  lambda ij: ij[0]):
    keys.append([j for i, j in ijs])
return keys
```
haven't tested it though

test/test-jaccard.py
```diff
@@ -0,0 +1,22 @@
+import unittest
```
erikbern: thanks for the unit test!

ann_benchmarks/distance.py
maumueller: Thanks, worked almost out-of-the-box. :-)

ann_benchmarks/datasets.py
maumueller: Sure! Added. (There are multipe `random_state=1` and `seed(1)` calls now. Maybe there should be a unique global seed.)

maumueller:
Hi Erik. What is the status on this? Is there anything you would like to discuss/see before merging?

erikbern:
sorry @maumueller i just forgot about it. feel free to ping me quicker next time in case a few days pass!

PR #133, opened 2019-11-29
Pynndescent update

lmcinnes:
It seems like pynndescent was broken (or otherwise drastically underperforming) in the last round. I've updated pynndescent, and pushed a new version to PyPI. This PR should update ann-benchmarks to use the new version (with it's changed parameters), and now only pull from PyPI (rather than master on the github repository) so it should be more stable in future.

### DIFF
--- a/algos.yaml
+++ b/algos.yaml
@@ -266,9 +266,14 @@ float:
       constructor: PyNNDescent
       base-args: ["@metric"]
       run-groups:
-        pynndescent:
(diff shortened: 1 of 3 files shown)


lmcinnes:
It looks like Puffinn and nmslib are failing in the tests, but I am pretty confident the changes in this pull request shouldn't touch either of those, so I'm not quite sure what the issue is there.

erikbern:
thanks! i will re-run benchmarks soon!

PR #137, opened 2020-01-02
On-Disk benchmarks

jpleet:
ANN is ideal for big data searches and it should be important to test cases when data is too big to store in memory. I've expanded several algorithms (annoy, faiss, ngt) and added a PostgreSQL approach (pgann) to benchmark on-disk (out-of-core) capabilities. Preliminary tests, though, use existing benchmark datasets.


erikbern:
sorry completely forgot about this. still somewhat interested in parts of this

PR #138, opened 2020-01-06
Support for jaccard with pynndescent

lmcinnes:
Pynndescent supports jaccard distances on datasets like kosarak via the scipy sparse matrix format. This updates the pynndescent algorithm wrapper to convert data into the requisite sparse format when using jaccard. I also pinned the numba and llvmlite versions as the current release combinations have some issues that break pynndescent.


erikbern:
nice! build was failing on some other algorithms (i'll try to fix one of these days)

PR #145, opened 2020-01-21
Add an dataset based on 'deep image descriptors'.

### LINKED ISSUE #143: Consider adding a 'deep image descriptor' dataset
In the past few years, it seems that the computer vision community transitioned away from representing images using traditional features like SIFT. It's now common to create 'deep descriptors' of images through convolutional neural nets (for example models based on the Inception architecture). It would be helpful to include a dataset of deep image descriptors to more closely represent this approach to image similarity search.

One idea is to take a sample from [this large dataset](http://sites.skoltech.ru/compvision/noimi/), which consists of images from ImageNet run through the a CNN to produce vectors. Some prominent kNN papers report results against this dataset, including [Billion-scale similarity search with GPUs](https://arxiv.org/abs/1702.08734) and [Efficient and robust approximate nearest neighbor search using Hierarchical Navigable Small World graphs](https://arxiv.org/abs/1603.09320).

jtibshirani:
This commit adds an image similarity dataset based on the data from 'Efficient
Indexing of Billion-Scale datasets of deep descriptors'. The data is composed of
images from the web that have been run through a 'deep descriptor network' to
produce vectors. The vectors were then reduced to 96 dimensions through PCA and
finally L2-normalized. We use a 10 million vector sample of the full dataset.

The original dataset can be downloaded from
http://sites.skoltech.ru/compvision/noimi/. The 10 million vectors are taken from
the file deep10M.fvecs. I emailed the authors of the dataset (with @erikbern in CC)
and they're okay with us committing a sample of the dataset.

Closes #143.

### DIFF
--- a/ann_benchmarks/datasets.py
+++ b/ann_benchmarks/datasets.py
@@ -211,9 +211,9 @@ def deep_image(out_fn):
     # length as an integer, then writing its components as floats.
     fv = numpy.fromfile(filename, dtype=numpy.float32)
     dim = fv.view(numpy.int32)[0]
-    X = fv.reshape(-1, dim + 1)[:, 1:]
+    fv = fv.reshape(-1, dim + 1)[:, 1:]
 
-    X_train, X_test = train_test_split(X)
+    X_train, X_test = train_test_split(fv)
     write_output(X_train, X_test, out_fn, 'angular')
 
 def transform_bag_of_words(filename, n_dimensions, out_fn):
--- a/ann_benchmarks/datasets.py
+++ b/ann_benchmarks/datasets.py
@@ -3,6 +3,8 @@ import numpy
 import os
 import random
 import sys
+
+from  urllib.request import urlopen
 try:
     from urllib import urlretrieve
 except ImportError:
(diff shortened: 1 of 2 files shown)


ann_benchmarks/datasets.py
```diff
@@ -192,6 +194,33 @@ def fashion_mnist(out_fn):
     test = _load_mnist_vectors('fashion-mnist-test.gz')
     write_output(train, test, out_fn, 'euclidean')
 
+# Creates a 'deep image descriptor' dataset using the 'deep10M.fvecs' sample
+# from http://sites.skoltech.ru/compvision/noimi/. The download logic is adapted
+# from the script https://github.com/arbabenko/GNOIMI/blob/master/downloadDeep1B.py.
+def deep_image(out_fn):
+    train_size = 10 * 1000 * 1000
+    test_size = 1000
```
erikbern: could we make the test set a bit bigger? 1k items make the recall numbers less reliable

erikbern: according to README.md this should be 10k also :)

ann_benchmarks/datasets.py
```diff
@@ -192,6 +194,33 @@ def fashion_mnist(out_fn):
...
+    # In the fvecs file format, each vector is stored by first writing its
+    # length as an integer, then writing its components as floats.
+    fv = numpy.fromfile(filename, dtype=numpy.float32)
+    dim = fv.view(numpy.int32)[0]
+
+    limit = (dim + 1) * (train_size + test_size)
+    fv = fv[0:limit]
+    X_train = fv.reshape(-1, dim + 1)[:, 1:]
+
+    X_train, X_test = train_test_split(X, test_size=test_size)
```
erikbern: the code currently fails on this line since `X` isn't defined. should L220 say `X` rather than `X_train`?

jtibshirani:
I'm sorry @erikbern for being sloppy here, I accidentally pushed a half-reverted refactor. The test set now has the correct size, and the dataset creation should run without error.

erikbern:
will take a look shortly. i tried to run this on my computer but it crashed because it was taking so much RAM :) might have to rent an EC2 instance

jtibshirani:
Great, thanks! I wanted to note there's a smaller sample file available called `deep10M.vecs` with a total of 10M vectors. It's ~3.6GB compared to the ~10GB file we're currently downloading. We could easily switch to using `deep10M.vecs` if it makes your workflow easier.

erikbern:
> Great, thanks! I wanted to note there's a smaller sample file available called `deep10M.vecs` with a total of 10M vectors. It's ~3.6GB compared to the ~10GB file we're currently downloading. We could easily switch to using `deep10M.vecs` if it makes your workflow easier.

If it's easy then I'd prefer that – I think more importantly it makes the choice of 10M vectors a bit less "arbitrary". But the faster download speed doesn't hurt!

jtibshirani:
I just updated to `deep10M.fvecs`. I initially thought it'd be nice to use the `base` files since it would easily allow for creating larger datasets in the future, but that might've been thinking too far ahead :)

erikbern:
this is great. i'll try to merge this weekend!

erikbern:
I was able to run this code and generate a dataset that I uploaded to http://ann-benchmarks.com/deep-image-96-angular.hdf5

Let's merge!

PR #149, opened 2020-02-17
Attempt at adding Milvus

erikbern:
A bit janky, since
(a) Milvus is a server process, so I hacked a hacky entrypoint file to run a daemon
(b) was impossible building it from source, but I realized I can just copy the relevant binaries from the "official" dockerfile and it seems to work

Running some benchmarks now, will see what we get

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
   - LIBRARY=faiss  DATASET=random-xs-20-angular
   - LIBRARY=flann DATASET=random-xs-20-angular
   - LIBRARY=kgraph DATASET=random-xs-20-angular
+  - LIBRARY=milvus DATASET=random-xs-20-angular
   - LIBRARY=mrpt DATASET=random-xs-20-angular
   - LIBRARY=n2 DATASET=random-xs-20-angular
   - LIBRARY=nearpy DATASET=random-xs-20-angular
--- a/algos.yaml
+++ b/algos.yaml
@@ -164,8 +164,8 @@ float:
       base-args: ["@metric"]
       run-groups:
         milvus:
-          args: [[1, 3, 10, 30, 100, 300, 1000]]
-          query-args: [[20, 40, 100, 200, 400, 1000, 2000, 4000, 10000, 20000, 40000]]
+          args: [['IVF_FLAT', 'IVF_SQ8'], [100, 300, 1000, 3000, 10000, 30000]]  # nlist
+          query-args: [[1, 3, 10, 30, 100, 300]]  # nprobe (should be <= nlist)
     nearpy:
       disabled: true
       docker-tag: ann-benchmarks-nearpy
--- a/algos.yaml
+++ b/algos.yaml
@@ -157,6 +157,15 @@ float:
           # This run group produces 3 algorithm instances -- Annoy("angular",
           # 100), Annoy("angular", 200), and Annoy("angular", 400) -- each of
           # which will be used to run 12 different queries.
+    milvus:
+      docker-tag: ann-benchmarks-milvus
+      module: ann_benchmarks.algorithms.milvus
+      constructor: Milvus
+      base-args: ["@metric"]
+      run-groups:
+        milvus:
+          args: [[1, 3, 10, 30, 100, 300, 1000]]
+          query-args: [[20, 40, 100, 200, 400, 1000, 2000, 4000, 10000, 20000, 40000]]
     nearpy:
       disabled: true
       docker-tag: ann-benchmarks-nearpy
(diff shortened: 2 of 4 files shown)


erikbern:
MNIST test:

image

Running Glove-25 right now

JinHai-CN:
Dear erikbern,

About milvus compilation, have you check following document ?
https://github.com/milvus-io/milvus/blob/master/INSTALL.md

erikbern:
@JinHai-CN yes, but I can't use the Milvus container off the shelf (or build the Docker container using the instructions). I need to add a bunch of stuff to make ann-benchmarks work.

erikbern:
Looks like it's capping out at about 30 queries per second. I'm wondering if it's some slowness caused by either (a) context switching (b) TCP delays (Nagle's algorithm)

I'm guessing batch queries would be faster, but ann-benchmarks is mostly built for serial queries

JinHai-CN:
Since it is a server, there are lots of config need to be set to make the query fast. I guess you didn't set the cache size to contain all data. Then, each query will make the data load from disk.

Here is the benchmark report we did on sift1b.
https://github.com/milvus-io/milvus/tree/master/docs

Back to ann-benchmarks, we would like to provide our contribution. I wonder if you can write which cases you want to test and we can give a guide to tuning Milvus on these cases.

RyanWei:
Hey Erik,

Seems you are testing our FLAT index type, you might want to try IVFSQ8 type which should be betterr.  
Also if GPU resource available, try IVFSQ8H which would use both GPU and CPU, should be the best.
For both index types, you might need to trigger the index creation manually since the test dataset might be too small to trigger auto indexing.  
Refer to: https://milvus.io/docs/v0.6.0/guides/milvus_operation.md#Create-an-index for more.

erikbern:
> Seems you are testing our FLAT index type, you might want to try IVFSQ8 type which should be betterr.

I was actually not using the FLAT index. I forgot to push the latest code to Github, but just did

erikbern:
> Back to ann-benchmarks, we would like to provide our contribution. I wonder if you can write which cases you want to test and we can give a guide to tuning Milvus on these cases.

Great if you can look at the current code and suggest any improvements!

JinHai-CN:
> > Back to ann-benchmarks, we would like to provide our contribution. I wonder if you can write which cases you want to test and we can give a guide to tuning Milvus on these cases.
> 
> Great if you can look at the current code and suggest any improvements!

Yes, we are checking the code and will provide some advices.

erikbern:
Looks like both datasets have an upper throughput of about 40 queries per second, which means about 1000/40 = 25ms per query.

What protocol are you using for the RPC? Is it HTTP or gRPC? I saw code for both in the python client

erikbern:
Any updates on this? I'd like to kick off a new benchmark soon, but I want to give you a chance to fix any issues with the Milvus code before that.

cxie:
Hi Erik,

I am Starlord from the Milvus team. Firstly, we would be happy to have Milvus listed in your new benchmark. We really appreciate all your time and efforts. Regarding Milvus benchmarking, we opened another PR, #153. We recommend you use the '"insert a vector with id" API for Milvus tests, as you did in other ANNS tests.

As a vector similarity search engine server, Milvus is designed for easy to use, high reliability, easy to scale, and high performance. Milvus supports near real-time search, CRUD, WAL, data consistency, distributed deployment, high availability, and much more. However, nothing good comes without a price. A fixed amount of server processing time, usually ~1 ms, will be added to each similarity search query, so the upper bond QPS of a single Milvus server instance would be around 1K. This was confirmed in both your benchmark results and our results (with some minor discrepancies, which I will discuss later). Milvus will perform much better in batch query mode and async single query mode because the ~1ms server processing time will be prorated among queries and become insignificant.

Back to Milvus benchmarking, we've run your ann-benchmarks code on an AWS EC2 c5.4xlarge instance to follow exactly the same route that you tested. We found out there are some discrepancies between your results and our results. Our plot of mnist-784-euclidean_10_euclidean results is capping out at about 770 QPS while yours is capping out at about 350 QPS. Moreover, our plot of glove-25-angular_10_angular results is capping out at about 900 QPS (very close to the theoretical upper bond mentioned above) while yours is capping out at about 400 QPS. We attached our plots and full results package for your further investigation.

Lastly, Milvus is positioned as an open source vector similarity search engine built on top of various ANNS algorithms. We are actively working on the support of more algorithms so that Milvus can better fit into different requirements and scenarios. We have planned to support Annoy and HNSW in Milvus 0.8.0 (April 2020). We may appreciate a new benchmark test on 0.8.0 after its release or when you think it's appropriate. 

mnist-784-euclidean_10_euclidean
glove-25-angular_10_angular
[milvus_EC2_c5.4xlarge_results.zip](https://github.com/erikbern/ann-benchmarks/files/4293183/milvus_EC2_c5.4xlarge_results.zip)

maumueller:
Hi Starlord

in case you didn't know: ann-benchmarks has support for batch mode, but I only added supported to it for a small set of implementations yet (scroll to the very bottom of http://ann-benchmarks.com/ to see some benchmarking results). 

You can use it by providing `--batch` and adding the correct call for it like in hnsw: https://github.com/erikbern/ann-benchmarks/blob/master/ann_benchmarks/algorithms/nmslib.py#L70-L71.

Might be worth a try. (And would be a reason to run more thorough experiments in batch mode.)

erikbern:
> Our plot of mnist-784-euclidean_10_euclidean results is capping out at about 770 QPS while yours is capping out at about 350 QPS.

The numbers I shared earlier was running on my laptop (not AWS) so that probably explains the difference!

Thanks for the other feedback, it's very helpful.

cxie:
@maumueller Thanks for the information. We will run the batch mode and share our results soon.

erikbern:
merging this now for #166

PR #151, opened 2020-02-27
specify cpu for each test-case(single querys mode)

tinkerlin:
Signed-off-by: Nicky <nicky.xj.lin@gmail.com>


erikbern:
thanks @tinkerlin

PR #153, opened 2020-03-05
Improve milvus.py

scsven:
Milvus provide a parameter named "ids" on the **insert** function that helps somebody want to specify id for every vector. This will be better than create an additional mapping.


erikbern:
Thanks!

Issue #130, opened 2019-09-17
Add N2 ?


vprelovac:
This one is promising:
https://github.com/kakao/n2

erikbern:
just glancing at the readme, definitely seems interesting.

@ummae – would you be interested in submitting a pull request to add n2 to the benchmarks?

ummae:
@erikbern Okay, I will.

erikbern:
great @ummae ! no rush

erikbern:
Ping @ummae do you want to contribute with it? I'd love to re-run the benchmarks soon

ummae:
@erikbern Sorry for late. This work is being delayed as we are preparing major update of N2. It will be updated in next 3-4 weeks (I hope so). It would be nice if anyone could contribute before then.

PR #154, opened 2020-03-10
add n2

taesaza0:
Sorry, we are late 🙏. We released the minor update and want to add it by changing the N2 configuration.
(Not yet major update, It seems to take longer than expected)

related issue: https://github.com/erikbern/ann-benchmarks/issues/130
related pr: https://github.com/erikbern/ann-benchmarks/pull/147

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Evaluated
 * [NGT](https://github.com/yahoojapan/NGT): ONNG, PANNG
 * [SPTAG](https://github.com/microsoft/SPTAG)
 * [PUFFINN](https://github.com/puffinn/puffinn)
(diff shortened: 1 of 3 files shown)


erikbern:
nice thanks!

erikbern:
i can kick off a new benchmark later this week – think it's overdue

taesaza0:
@erikbern 
I understand that there is only one cpu in the benchmark, but I wonder if it's the same in the build phase(ex, hnsw build)
We explicitly set the number of threads to one for build and search phase, is it the right way?

erikbern:
it's possible we can use multiple cpus for the build phase, but i think it's fine that it's a bit slower than it could be

Issue #104, opened 2018-12-21
for help: code raise a exception


CogmanCVN:
when runing install.py this line of code raise a exception，
subprocess.check_call( 'docker build \   --rm -t ann-benchmarks -f install/Dockerfile .', shell=True) 
Exception is：
subprocess.CalledProcessError: Command 'docker build         --rm -t ann-benchmarks -f install/Dockerfile .' returned non-zero exit status 1.
I hope you can help me

appusom:
Is there any solution to this problem? I am facing the same on Ubuntu 16.04.

erikbern:
Does it print a bunch of stuff to standard out? Can you share?

appusom:
I finally resolved this issue by ensuring the docker daemon was running correctly before I ran this code. Thanks.

pramodv1993:
Hello, I too am facing the same issue once I try to run the install.py script, Is there a solution for the same?

maumueller:
What does it say if you run ` docker build --rm -t ann-benchmarks -f install/Dockerfile .` in your terminal (in the ann-benchmarks directory)?

yurymalkov:
I also has experienced errors with the docker build, probably due to ubuntu update.
Changing the first line of the docker file to `FROM ubuntu:18.04` helped me.

pramodv1993:
> What does it say if you run ` docker build --rm -t ann-benchmarks -f install/Dockerfile .` in your terminal (in the ann-benchmarks directory)?

Hello Sir, I tried the suggestion by Yury Malkov, it now passed the step6/7 of  RUN pip3 install -rrequirements.txt :)

pramodv1993:
> I also has experienced errors with the docker build, probably due to ubuntu update.
> Changing the first line of the docker file to `FROM ubuntu:18.04` helped me.
Hello Sir, this helped pass the step where I had got stuck. Thank you :)

PR #161, opened 2020-05-28
fix ubuntu version

### LINKED ISSUE #160: List of tested platforms
Do you have a list of platforms that install.py has been tested and verified on?

I've tried running on both OSX catalina and a fresh installation of Ubuntu 18.04, each with docker installed. Installation fails with errors on both platforms.

yurymalkov:
should fix #160 (and part of #104)


erikbern:
thanks!

PR #165, opened 2020-06-25
Add ScaNN to ann-benchmarks

sammymax:


sammymax:
Attached are our preliminary results on Glove-100-angular, k=10. They are single-threaded, non-batched results run on a Xeon W-2135. We are planning on using these results in our paper _Accelerating Large Scale Inference with Anisotropic Vector Quantization_ (recently accepted to ICML 2020) but naturally we also want our algorithm to appear in the official comparison as well.

results

erikbern:
wow, impressive results!

do you mind adding it to .travis.yml as well so that it runs as a part of CI? 

i _really_ need to re-run the whole suite of algorithms and publish some new results. will try to find some time in the next few weeks

sammymax:
Done!

erikbern:
nice! thanks for the contribution!

erikbern:
@sammymax my understanding is scann supports euclidean distance as well. is it as easy as removing the normalization `X /= np.linalg.norm(X, axis=1)[:, np.newaxis]`? if so, I can add that to the benchmark suite

sammymax:
That's the most important change, but there are also a few parameter choices that don't make sense or aren't supported with Euclidean distance that have to be changed. I'll try to get a configuration for Euclidean running in the next few weeks.

erikbern:
Ok – I'm currently running a benchmark right now with the plan to publish the results in about a week, so if it's easy to get it done by then, I could include ScANN for Euclidean as well. It's currently SotA for angular, would be great to get it included for Euclidean as well!

sammymax:
Unfortunately, I got busy with a few other things, and given there's four Euclidean datasets of very different sizes, dimensionalities, and intrinsic dimensionalities (some of these datasets' idiosyncrasies are discussed in the last page of [our paper](https://arxiv.org/pdf/1908.10396.pdf)) I don't think I'll be able to get tuned configurations in time. Next time, though!

erikbern:
Ok no problem. I think there’s enough reason to re-run the benchmarks in a couple of months. Hopefully there will be something by then :)

PR #168, opened 2020-07-01
make it possible to run multiple containers in parallel

erikbern:
This should do the trick :)

Running it on MNIST right now to see if it works

### DIFF
--- a/ann_benchmarks/main.py
+++ b/ann_benchmarks/main.py
@@ -30,6 +30,17 @@ def positive_int(s):
     return i
 
 
+def run_worker(cpu, args, queue):
+    while not queue.empty():
+        definition = queue.get()
+        if args.local:
+            run(definition, args.dataset, args.count, args.runs, args.batch)
+        else:
+            mem_limit = int(psutil.virtual_memory().available / args.parallelism)
+            run_docker(definition, args.dataset, args.count,
+                       args.runs, args.timeout, args.batch, str(cpu), mem_limit)
+
+
 def main():
     parser = argparse.ArgumentParser(
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
@@ -98,11 +109,6 @@ def main():
         '--run-disabled',
         help='run algorithms that are disabled in algos.yml',
         action='store_true')
-    parser.add_argument(
-        '--cpu-number',
-        type=positive_int,
-        help='specify cpu number',
-        default=0)
     parser.add_argument(
         '--parallelism',
         type=positive_int,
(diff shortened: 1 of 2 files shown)


maumueller:
I would be interested in the results of 8x annoy in parallel vs. a single-threaded version on a 8 vCPU ec2 instance. I don't know how much load the ec2 instances have otherwise, but 8x times avx2 distance computations might show some serious performance drops in query time.

erikbern:
I don't think this change should have any impact on speed right? Each container is limited to 1 CPU, so they will get their own CPU (as long as the parallelism is lower than the number of CPUs, of course)

That being said, I just noticed running `top` that currently each process has exactly 25% CPU (there are 4 of them) so I think I messed up something with how the CPUs are determined. Let me fix!

maumueller:
I don't know which CPUs are used in the current ec2 generation, but you may always run into problems with shared cache lines and clock downscaling if there are too many heavy SIMD instructions. It's best to benchmark that :-)

erikbern:
It's true you might seem some more cache thrashing with higher parallelism. Idk feels like it's worth it though in order to bring down the total runtime by 4-8x. It should affect all the algorithms approximately the same anyway. Maybe we can increase the number of runs to more than 2.

I think something like 98% of all time is spent building the index as opposed to running the queries, so it's unlikely that more than one algorithm is running queries at any point in time. Running the queries 3-5 times would warm up the cache on that CPU.

erikbern:
Rewrote this to actually farm it out to each CPU. I think this should work. I wiped the results for MNIST and I'm rerunning it now. Just looking at the output it looks _much_ faster, and when I run `top` I see 7 processes running at 100% CPU each.

If this works out well them I'm tempted to re-run the glove benchmarks using this as well (but maybe using say 3 or 4 processes not 7).

erikbern:
I can also run the full benchmark on a smaller random dataset and see if there's any difference in the results with/without this change.

maumueller:
As I said above, I think annoy on MNIST might already show a performance issue if there is one. 

Maybe compare single-threaded/4 parallel processes/8 parallel processes/16 parallel processes (the latter just as a sanity check to see that it performs much worse).

erikbern:
Sure, I can do that. It assigns CPUs 1...args.parallelism so I can't do more than 15 on a c5.4xlarge. I'm currently running with 7. Once it's done let me try with 1 and see what's the difference.

erikbern:
With 7 in parallelism (whole run took <2h)

image

Running with 1 now to see. It will take probably more like 24h.

erikbern:
... and here's the benchmark with parallelism = 1 (it took almost 24h to run as I expected).

image

There _is_ a small different, but it seems to be no more than 5-10%, and more importantly, it doesn't change the relative ranking (seems to affect all algorithms equally).

My thinking is to merge this, increase the number of runs from 2 to 5, and then run all benchmarks with parallelism 3-5 in order to finish everything _much_ faster.

erikbern:
Merging this for now and kicking of a glove build

PR #169, opened 2020-07-13
New run July 2020

erikbern:
I messed up with Git so I had to force-push to master to get it back to it previous state. Think it should be good now!


erikbern:
Results here: http://ann-benchmarks.com/results-2020-07-12/ @maumueller 

I'll post it to the main site once this PR gets merged.

I'm tempted to rerun this pretty soon so I don't want to make a huge deal out of it.

erikbern:
Some more notes:
* The gist-960 results are incomplete. The benchmark is so slow that I gave up after a few days. This is the only one that I can't run with parallelism but I can do that if I get a bigger machine.
* Scann does exceptionally well on glove-100-angular but not at all on glove-25-angular and it fails on nytimes-256-angular (and I haven't had time to check why). I believe the current code is somewhat experimental so I'd love to see some updates
* Milvus doesn't seem very strong, but I believe there's something slowing its queries down and hopefully it can be fixed.

yurymalkov:
Scann does not work on nytime due to #73, which causes an error in normalization.
I've also tested scann. It works very great on glove, but loses to other algorithms on deep10m (the dataset seem already present  https://github.com/erikbern/ann-benchmarks/blob/master/ann_benchmarks/datasets.py#L194). 
Scann seems to work best for the case for degenerate data (high dim at small datasets, which seem to make sense), but it does not scale well when the dataset size increases or the dimensionality goes down.

erikbern:
Merging this tomorrow and pushing new results to the website unless anyone objects!

masajiro:
Thank you for running the benchmarks.
Do you have any idea or any error messages about why there are no results of NGT in the glove-100-angular and gist-960-euclidean? Is there any possibility that running the benchmarks in parallel caused it?

erikbern:
@masajiro I don't have the error message, but most likely it ran out of memory running those. Each container had about 10GB of memory allocated to it.

yurymalkov:
In GIST only the data takes 7.5 Gb (one copy for the benchmark, one for the algorithm; pynndescent seem to use the initial data), that leaves only 2.5Gb for the ann index and everything else. 
I wonder, why not allow multi-threaded construction (with control of the queries by changing the process cpu affinity)?

erikbern:
@yurymalkov I ran the GIST benchmark with no parallelism so there would be more memory (30GB). But after several days it was still only 20-25% through and so I gave up. 

I've thought about using multiple threads for index construction, but it's a bit complicated to implement since CPU affinity is controlled from outside the container.

My thinking is next time I'll use eg a r6g.4xlarge instead – they have 16 CPUs and 128GB RAM. That way I could run the benchmarks with much higher parallelism, eg 10 simultaneous runs each getting 12GB memory. At that point using multiple threads isn't really that meaningful since I'd already saturate 10/16 CPUs

masajiro:
I think you can get at least a result of NGT-onng for glove-100-angular by running without parallelism. Since I was able to produce results of NGT for gist-960-euclidean on c5.4xlarge, I am wondering why NGT cannot work well on your server.

erikbern:
> I think you can get at least a result of NGT-onng for glove-100-angular by running without parallelism. Since I was able to produce results of NGT for gist-960-euclidean on c5.4xlarge, I am wondering why NGT cannot work well on your server.

That's probably true, but it's really slow to run without parallelism. It would have taken probably a week to run all the algorithms without any parallelism, versus just 2 days with (using 3 workers)

yurymalkov:
@erikbern I see. I think calling os.sched_setaffinity() inside the script should limit the number of visible cpus during the queries, although I haven't checked that it is actually working.

erikbern:
Merging this with the expectation that I'll re-run the benchmark in the next couple of months!

gony0:
@erikbern The date doesn't seem to have been updated!
[These are all as of 2019-06-10, run...](https://github.com/erikbern/ann-benchmarks/blob/master/README.md#results)

erikbern:
> @erikbern The date doesn't seem to have been updated!
> [These are all as of 2019-06-10, run...](https://github.com/erikbern/ann-benchmarks/blob/master/README.md#results)

oops, that's just the date that wasn't updated. I'll fix!

PR #172, opened 2020-07-16
Tune ScaNN for other angular datasets

sammymax:
The originally submitted configuration was only tuned for Glove-100. Here are some better configurations for the other angular datasets. Still investigating NYTimes...

Glove-25:
glove-25
LastFM:
lastfm


erikbern:
Nice!

FYI nytimes-256 has a few "missing" vectors (all elements set to zero) which I guess is a bug or a feature depending on how you look at it (I've been arguing that's a common case that libraries _should_ ideally be able to handle). So that might cause issues for ScANN

erikbern:
Let me know if you want me to merge this. Otherwise will keep it open so you can optimize more :)

sammymax:
Sure, please merge, thanks!

PR #175, opened 2020-08-10
add ONNG parameters for the last.fm

masajiro:
The parameters will probably improve the ONNG result for the last.fm dataset.


erikbern:
thanks!

PR #179, opened 2020-09-16
Update PyNNDescent

lmcinnes:
I recently released a new version of pynndescent. This should get ann-benchmarks working with it. There are a couple of other changes from issues encountered with my own testing. First the addition of ``freeze_support`` to make things play nice with newer multiprocessing releases, and second a minor change to the ScaNN model so that it will at least run on NYTimes (the all zero vectors in that dataset were breaking it).


erikbern:
thanks! just checking, the amount of runs is still reasonable right? i don't want the benchmark suite to blow up :)

lmcinnes:
It is actually less runs; just broken up differently, so it expands out in the algos.yaml.

Right now I'm fighting with some issues of Docker (I run things with --local for testing). Hopefully I'll get those sorted soon.

lmcinnes:
Out of curiousity, does the jaccard benchmark get run at all? I have been trying to maintain the code to make it work, but I've never seen benchmarks posted for Kosarak, so I'm wondering if it is worthwhile.

erikbern:
I haven't looked at Jaccard – it was something @maumueller built. I can try to run it next time and see how many algorithms support it.

lmcinnes:
Looks like it is just faiss and milvus now, which I don't believe I touched. Thanks again for taking the time to build all of this. It has been invaluable for me to have a consistent way to do my own benchmarking, and it provides a great service to the community on the current state-of-the-art.

lmcinnes:
It seems like milvus just has a stale link for a config file (although the version also seems very out of date; it is 0.6.0, but they seem to be on 0.10.2 now). I'm not sure where to get the right one though. Faiss seems to be due to the fact that they have moved from autotools to cmake for build configuration. Ideally that isn't too hard to fix but they seem to have a lot of carefully crafted flags in their install file, so I wouldn't really want to mess with that. I don't believe I can fix either of those, so this is about as good a state as this PR is ever going to be in I'm afraid.

erikbern:
No worries, not your job to fix those other algorithms!

stephenleo:
Can we merge this in? build is still failing at pynndescent. At least one of the issue is with the renaming of a parameter `diversify_epsilon` to `diversify_prob` which this PR seems to fix.

If you need some help to test it on Debian, Ubuntu, etc let me know. I can run it on a fresh GCP VM.

erikbern:
looks like this one is now going to have to be rebased :/

lmcinnes:
Conflicts should be resolved now.

erikbern:
@lmcinnes I just ran this locally but it's failing at the last step:

```
Step 7/7 : RUN python3 -c 'import pynndescent'
 ---> Running in ecc021cc872b
double free or corruption (top)
Aborted
The command '/bin/sh -c python3 -c 'import pynndescent'' returned a non-zero code: 134
```

erikbern:
Thanks @lmcinnes !

erikbern:
Unfortunately it's failing after merging:

```
Step 7/7 : RUN python3 -c 'import pynndescent'
 ---> Running in eb006d9d49ec
double free or corruption (top)
Aborted
The command '/bin/sh -c python3 -c 'import pynndescent'' returned a non-zero code: 134
```

erikbern:
Fix in #198

PR #180, opened 2020-09-17
Elastiknn for Euclidean datasets

alexklibisz:
First pass at setting up Elastiknn. 

For now I'm only including it for Euclidean datasets. I'll work on angular, jaccard and hamming over the next couple weeks.

I tested this out on a C5.4xLarge and it seems to work reliably with `--runs 5 --parallelism 3 --count 100`. I'll run it again over night tonight to be extra sure it runs to completion. It is quite a bit slower than the existing algos, for pretty obvious reasons (Java, making HTTP requests, writing to disk, etc.). The numbers I was seeing in my ann-benchmarks runs are pretty much the same as I saw in my own benchmarking, [which is reported here](http://elastiknn.klibisz.com/performance/). Fashion mnist peaks at about 200 q/sec and sift at about 100 q/sec.

I made a slight modification to the runner, so that when a container fails, it prints the short ID of the container. This was helpful in debugging, but I'm also happy to revert it.

LMK if there's anything I'm missing in the integration and I'll fix it ASAP. The only thing I'm fuzzy on: how does travis know to build the image? I was manually running `docker build` commands locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ env:
   - LIBRARY=mih DATASET=random-xs-16-hamming
   - LIBRARY=datasketch DATASET=random-s-jaccard
   - LIBRARY=scann DATASET=random-xs-20-angular
+  - LIBRARY=elastiknn DATASET=random-xs-20-angular
 
 before_install:
   - pip install -r requirements.txt
--- a/algos.yaml
+++ b/algos.yaml
@@ -218,7 +218,7 @@ float:
         M-96:
           arg-groups:
             - {"M": 96,  "efConstruction": 500}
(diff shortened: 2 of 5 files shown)


alexklibisz:
FYI, the current setup timed out last night. Trying it again with fewer query args.

alexklibisz:
This is still timing out on the sift dataset. I could have sworn I had run it a couple times to completion before making the PR, but maybe I was actually running the fashion mnist benchmark. I need to do some profiling and will re-open once it's back in shape.

erikbern:
This looks pretty good! Let me know once you're able to fix the timeout issue

alexklibisz:
@erikbern Do you run the "official" benchmark using batch mode or single queries?

erikbern:
@alexklibisz single queries

PR #183, opened 2020-10-06
Update faiss dockerfile

strelok2012:
Faiss changed build system to Cmake instead of configure, and current dockerfile doesn't work


erikbern:
awesome, thanks!

PR #196, opened 2020-12-04
closes #194. Also adds some resilience to failing packages

### LINKED ISSUE #194: Default Colors are a bit hard to see on certain algorithms
The default plot colors are too light to clearly see on a white background (especially for faiss-ivf, hnsw(nmslib). I propose a slight modification to the colors to make it more readable in my opinion. What do you folk think?

Code change to ann-benchmarks/plotting/utils.py:
```
def generate_n_colors(n):
    vs = numpy.linspace(0.3, 0.9, 7)
```

Defaults:
image

Updates:
image

stephenleo:
This PR addresses 2 things:
1. The colors in the plot are made slightly darker as reported in #194 
2. Individual algorithms that error while installing will not stop the installation of other algorithms. Instead, we will install whatever algorithm does install and print the install status of all algorithms at the end.

### DIFF
--- a/ann_benchmarks/plotting/utils.py
+++ b/ann_benchmarks/plotting/utils.py
@@ -87,7 +87,7 @@ def compute_all_metrics(true_nn_distances, run, properties, recompute=False):
 
 
 def generate_n_colors(n):
-    vs = numpy.linspace(0.4, 1.0, 7)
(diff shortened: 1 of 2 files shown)


erikbern:
thanks!

PR #197, opened 2020-12-05
Fix build error

erikbern:
Should resolve current build problems


erikbern:
not sure why travis stopped running on branches?

PR #200, opened 2020-12-06
fix back separation between batch and non batch

erikbern:
i was a bit too eager to remove code, see https://github.com/erikbern/ann-benchmarks/commit/cc5d20afa0167bf2000def219f826e18e5eb7184


erikbern:
thanks for pointing it out, @maumueller!

PR #201, opened 2020-12-07
simplify the batch stuff further

erikbern:
my previous attempts to simplify the results code wrt batching introduced a bunch of issues that i guess i could have reverted, but instead i decided to keep digging a bit and i think i was able to simplify it even further and get things working again

1. batching is no longer a part of any algorithm name, just the filename
2. we do no longer infer whether batching is used from the filename, but instead read it from the result metadata

i think it's a bit cleaner

### DIFF
--- a/ann_benchmarks/algorithms/definitions.py
+++ b/ann_benchmarks/algorithms/definitions.py
@@ -16,12 +16,6 @@ Definition = collections.namedtuple(
      'arguments', 'query_argument_groups', 'disabled'])
 
 
-def get_algorithm_name(name, batch):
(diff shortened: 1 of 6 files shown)


erikbern:
I'll wait a few days to merge just in case @maumueller has anything to say. If this breaks anything, I'll just revert all my latest changes to this back to what they were before.

erikbern:
travis is messed up but https://travis-ci.org/github/erikbern/ann-benchmarks/builds/748148360

erikbern:
merging it so that master can be green

PR #205, opened 2020-12-17
Fix batch mode behavior

maumueller:
A follow-up on #201. I think the batch mode file handling works nicely again (except #203). 

However, the whole docker parallelization rendered batch mode in Docker useless and this PR fixes this. Batch mode was meant to allow an implementation to go 'all in' with CPU/GPU resources. It doesn't help an implementation much if it receives all the queries at once if it then has to solve it in a single thread :-) 

I tried to find a fix that was minimally invasive with the rest of the code base. Let me know what you think.

### DIFF
--- a/ann_benchmarks/algorithms/base.py
+++ b/ann_benchmarks/algorithms/base.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+from multiprocessing.pool import ThreadPool
 import psutil
 
 
@@ -19,9 +20,11 @@ class BaseANN(object):
         return []  # array of candidate indices
 
     def batch_query(self, X, n):
-        self.res = []
-        for q in X:
-            self.res.append(self.query(q, n))
+        """Provide all queries at once and let algorithm figure out
+           how to handle it. Default implementation uses a ThreadPool
+           to parallelize query processing."""
+        pool = ThreadPool()
+        self.res = pool.map(lambda q: self.query(q, n), X)
 
     def get_batch_results(self):
         return self.res
--- a/ann_benchmarks/main.py
+++ b/ann_benchmarks/main.py
@@ -41,8 +41,12 @@ def run_worker(cpu, args, queue):
         else:
             memory_margin = 500e6  # reserve some extra memory for misc stuff
             mem_limit = int((psutil.virtual_memory().available - memory_margin) / args.parallelism)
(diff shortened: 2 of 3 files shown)


erikbern:
I thought the point of batch was for algorithms that use minibatches on the GPU or similar :)

But I don't have any strong feelings about it. However – if the point is to go all in on the GPU/CPU then I wonder if it would make sense to change the default implementation of batch queries to use a thread pool when there's no custom batch query interface?

maumueller:
Nice suggestions, implemented it. Gave a speedup of roughly 3 in QPS for annoy on my 4-core laptop. Noticed another problem in the website creation that I fixed as well.

ann_benchmarks/algorithms/base.py
```diff
@@ -19,9 +19,12 @@ def query(self, q, n):
         return []  # array of candidate indices
 
     def batch_query(self, X, n):
-        self.res = []
-        for q in X:
-            self.res.append(self.query(q, n))
+        """Provide all queries at once and let algorithm figure out
```
erikbern, line 22: isn't there code for

ann_benchmarks/algorithms/base.py
```diff
@@ -19,9 +19,12 @@ def query(self, q, n):
         return []  # array of candidate indices
 
     def batch_query(self, X, n):
-        self.res = []
-        for q in X:
-            self.res.append(self.query(q, n))
+        """Provide all queries at once and let algorithm figure out 
+           how to handle it. Default implementation uses a ThreadPool
+           to parallelize query processing."""
+        from multiprocessing.pool import ThreadPool
```
erikbern, line 25: super minor but i'd have a preference for this at the top module level

erikbern:
Looks great, is this ready to be merged? Happy to do so whenever

ann_benchmarks/algorithms/base.py
maumueller: Fixed.

maumueller:
Sorry, lost track of this. I squashed the commits and rebased. Ready to be merged. 

(I was about to sneak in adding an explicit pointer to the interactive visualisations from the plots in `README.md`, but maybe this should be its own PR. :-))

erikbern:
sorry dropped the ball on this (along with some other ann-benchmarks stuff), but merged it!

PR #206, opened 2020-12-21
Fix result caching

sammymax:
Result files end in .hdf5 but the current code neglects this, so it never finds the file, even if the config has already been run.

### DIFF
--- a/ann_benchmarks/main.py
+++ b/ann_benchmarks/main.py
@@ -155,7 +155,7 @@ def main():
         for query_arguments in query_argument_groups:
             fn = get_result_filename(args.dataset,
                                      args.count, definition,
-                                     query_arguments, args.batch)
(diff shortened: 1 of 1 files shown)


erikbern:
This seems related to a refactoring I did a few weeks ago. I don't know if this is the right place to fix it though (ideally it would be fixed in the `get_results_filename` helper). I'll take a look. Thanks for flagging!

sammymax:
No problem, feel free to fix this however you think is best (and disregard/close this PR if needed)!

erikbern:
sorry, dropped the ball, hopefully i can take a look at it this week!

maumueller:
This has been taken care of inside of `get_result_filename` in ae0f175b811f4d95e1f14f4894d2e1eeb5884a3a.

PR #208, opened 2021-01-09
Added documentation in run_algorithm.py

thomasahle:
This documents the arguments taken by `run_algorithm.py` and adds a big notice to the `--help` saying the user probably wants to run `run.py` instead.


erikbern:
Seems reasonable, thanks!

PR #211, opened 2021-01-14
Update ScaNN to support Euclidean datasets

sammymax:
There's also some other minor changes, like adding ThinLTO to the build flags and adding a tuning better suited to the NYTimes dataset, but those tweaks are pretty marginal.


erikbern:
Nice! I'll see if I can rerun the benchmarks soon

PR #214, opened 2021-01-26
Updated faiss docker file

maumueller:
All FAISS implementations turned out pretty slow in the last benchmark run. This PR fixes that.


erikbern:
Thanks!

PR #215, opened 2021-01-29
Stop container after timeout.

maumueller:
7d767f53210981fac8fe2bfb559ce1410b1a1af0 sneaked in a bug so that it doesn't stop containers after running into a timeout. This had the nice effect that some containers would later share the CPU with a container that was supposed to be killed.


maumueller:
Any reasons not to merge this @erikbern?

erikbern:
oh sorry, not sure why i dropped the ball on this one. thanks!

PR #213, opened 2021-01-18
More scales for plotting

thomasahle:
For https://github.com/erikbern/ann-benchmarks/issues/209

PR #216, opened 2021-02-01
More parameters for scann

thomasahle:
The smallest number of buckets probed by ScaNN is 30. This gives the wrong impression that ScaNN has a much better space partitioning than the remaining algorithms when it rather is just very fast during simple scans. This commit adds a few more data points, which also makes ScaNN look more like the remaining algorithms in plots.


sammymax:
We've never tuned ScaNN for low recall because our impression is very few people are interested in such low recalls. ann-benchmarks shows the whole 0% to 100% recall range with equal weight (although maybe your #213 changes things?) so I understand if you want to submit low-recall tunings.

Are these properly tuned though? It doesn't make sense to do reordering of 10 elements just to return the top 10 of them. Also, what is the wrong impression that ScaNN currently gives?

thomasahle:
@sammymax Maybe you are right that the reordering numbers are not perfect. They give pretty good results in my experiments though.

Because the ScaNN plot starts at recall around 0.8, I at least, for a long time assumed this was the recall when probing just a single bucket. In other words, it seemed like ScaNN had some amazing space partition based on your new anisotropic ideas, that I spent a long time trying to understand.

In any case, showing the results for smaller recalls seems to be the standard for ann-benchmarks.

sammymax:
Got it--given how much space ann-benchmarks devotes to low recalls, I guess some more tuning points in that region makes sense.

Would you mind raising all the 10's to like 20 or 30? I haven't really tested but my hunch would be that you'd get way better recall for basically no QPS cost.

thomasahle:
That makes a lot of sense. I have updated the patch

erikbern:
Thanks!

PR #220, opened 2021-02-23
Fix install milvus failed.

lajiyuan:
Build milvus by `python install.py --algorithm milvus ` repeatedly will be failed, because that  docker sees the initial and modified instructions as identical and reuses the cache from previous steps. As a result the apt-get update is not executed because the build uses the cached version. Because the apt-get update is not run.

For details, please refer to this article: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/


lajiyuan:
Screenshot of compile image error：

erikbern:
Thanks – the Milvus is currently a very old version, so we should update it at some point

PR #221, opened 2021-03-01
Update NGT parameters

masajiro:
To shorten the benchmarking time, I updated the parameters for ONNG and QG of NGT.
Thank you in advance.


erikbern:
Thanks!!

ShikharJ:
@erikbern I presume the update to the plots was unintentional? Or can we get DiskANN runs added to these plots as well?

masajiro:
I unintentionally added the results of our test.

maumueller:
I'll merge the revert in a second, but maybe it's a good time to rerun the benchmark, @erikbern? scann has euclidean distance support, diskann was added and seems to perform very well, and onng also has some new tricks :-) 

I don't know at which point we lowered the timelimit from 5 to 2 hours, but I believe @erikbern had a good reason for doing that. Would it be possible for you, @masajiro, to suggest parameter values that support building the graph in time?

erikbern:
Yeah I'd be open to rerunning the benchmarks for sure

masajiro:
@maumueller 
Thank you for asking. The 5 hour timeout value is better for building ONNG. Although ONNG parameters might be adjusted for the 2 hour value, its performance could decrease much. If the long benchmarking time is an issue, I think that it is effective to allow building processes to use multiple cores.

Issue #225, opened 2021-03-23
python3 create_website.py to generate website， ERROR 'Permission denied'


lajiyuan:
Should we execute "chmod -R 777 results" after writing out the result folder every time, otherwise we will not have read permission and we cannot use create_website.py to generate a website？

maumueller:
The idea is to to run `create_website.py` with `sudo`. The docker container writes the result files as `root`, and `create_website.py` will cache some computations and needs write access to these files. @erikbern Maybe we should run the docker images as the actual user.

erikbern:
I think Docker generally likes to run as the superuser on a system, and I'm not sure if we have control over that, but I could be wrong?

My preference here would be to not mess around with permissions if possible. Maybe it's possible to write the website to a different directory or to mount a subdirectory of the results/ directory?

maumueller:
I think it's pretty straight-forward to do, see e.g., https://github.com/lemire/docker_programming_station/blob/master/Dockerfile#L21-L27, from https://lemire.me/blog/2020/05/22/programming-inside-a-container/. 

The problem is not with the website creation, but with the `hdf5` files being owned by `root`, and the metrics scripts trying to cache metrics by writing them into the `hdf5` file.

erikbern:
If it can be fixed in the Dockerfile, let's definitely do that. I didn't realize it was possible.

PR #226, opened 2021-03-24
[WIP] Run docker as user

maumueller:
As discussed in #225. 

Let's see what CI says to that.

### DIFF
--- a/install.py
+++ b/install.py
@@ -12,7 +12,7 @@ def build(library, args):
         q = " ".join(["--build-arg " + x.replace(" ", "\\ ") for x in args])
     else:
         q = ""
-    
+
     try:
         subprocess.check_call(
             'docker build %s --rm -t ann-benchmarks-%s -f'
@@ -48,7 +48,9 @@ if __name__ == "__main__":
     print('Building base image...')
     subprocess.check_call(
         'docker build \
-        --rm -t ann-benchmarks -f install/Dockerfile .', shell=True)
+        --rm -t ann-benchmarks -f install/Dockerfile \
+         --build-arg USER_NAME="$USER"  --build-arg USER_ID=$(id -u) \
+         --build-arg GROUP_ID=$(id -g) .', shell=True)
 
     if args.algorithm:
         tags = [args.algorithm]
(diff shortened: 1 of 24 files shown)


alexklibisz:
Maybe it would be simpler to execute `sudo chown -R $(whoami):$(whoami) results` or equivalent `chmod` from the python runner? I think I've solved this issue at some point before (I remember it was a pain), so if I remember where and how I'll post back here.

maumueller:
Thanks for the comment @alexklibisz . I didn't look further into this, but it seems to be working with the exception of `milvus` (ping @scsven: can you see the problem with https://travis-ci.org/github/erikbern/ann-benchmarks/jobs/764372092?)

I'm a bit unsure how your approach of  `chown`ing in the python runner would actually work: How would you propagate the username? I also think it's difficult to get the group and user ids of the container and the host matched up. Would be interesting to see a working version of your approach!

scsven:
@maumueller Please make sure Milvus running as root in the container. I ran your branch on my machine, the errors on Travis-ci can be reproduced. When I add `sudo LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/var/lib/milvus/lib` in install/Dockerfile.milvus:22, the errors dismissed.

Whole line:
`RUN echo 'sudo LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/var/lib/milvus/lib /var/lib/milvus/bin/milvus_server -d -c /var/lib/milvus/conf/server_config.yaml -l /var/lib/milvus/conf/log_config.conf' >> entrypoint.sh`

erikbern:
Sorry completely missed this PR. Is it necessary to do it this way? If we want to solve #225 can't we just run the website creation as a user? alternatively the superuser could just `chmod a+r`

maumueller:
I think this is not just about #225 but more general about creating files with the correct rights. @scsven's suggestion is running through CI; then it should be working.

I'm honestly not sure as well: It's kind of nice that it creates the files with the correct user and group, and I actually think it's an advantage to see the `sudo` in case users see the dockerfile as a reference installation guide.

Giving a+w (the problem is writing to the file for caching of results) at some well defined point would work as well. It probably has to be done within the Python runner with the current Docker setup, which is pretty strange.

erikbern:
fair enough, makes sense

is this ready to be merged?

maumueller:
Let me smash the commits and do some more test runs, I'll get back to you on this.

PR #230, opened 2021-03-29
Add Vamana and Vamana-PQ (DiskANN) to ANN Benchmarks

ShikharJ:
Hey @erikbern and team,

We've added the support for Vamana and Vamana-PQ in-memory indices (from the [DiskANN paper](https://proceedings.neurips.cc/paper/2019/hash/09853c7fb1d3f8ee67a61b6bf4a7f8e6-Abstract.html)) to your amazing repository. Correspondingly, we've run a few comparison experiments with some other top algorithms on our Azure VM (Intel(R) Xeon(R) Platinum 8272CL CPU + 128 GB RAM). Both the variants seem to offer competitive performance under different settings.

fashion-mnist-784-euclidean
gist-960-euclidean
glove-25-angular
glove-100-angular
lastfm-64-dot
mnist-784-euclidean
nytimes-256-angular
sift-128-euclidean

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ env:
   - LIBRARY=elasticsearch DATASET=random-xs-20-angular
   - LIBRARY=elastiknn DATASET=random-xs-20-angular
   - LIBRARY=opendistroknn DATASET=random-xs-20-angular
+  - LIBRARY=diskann DATASET=random-xs-20-angular
 
 before_install:
   - pip install -r requirements.txt
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ Evaluated
 * [ScaNN](https://github.com/google-research/google-research/tree/master/scann)
 * [Elastiknn](https://github.com/alexklibisz/elastiknn)
 * [OpenDistro Elasticsearch KNN](https://github.com/opendistro-for-elasticsearch/k-NN)
+* [DiskANN](https://github.com/microsoft/diskann): Vamana, Vamana-PQ
 
 Data sets
 =========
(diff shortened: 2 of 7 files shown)


ShikharJ:
Ping @erikbern

maumueller:
Looks great @ShikharJ! Just out of curiosity: What is the difference between intel mkl and openblas? :-) All the `FAISS` code uses openblas at the moment, and I don't know what `scann` uses.

ShikharJ:
@maumueller I'm not sure if we ever ran comparison tests for MKL vs OpenBLAS. My understanding is that since OpenBLAS doesn't yet support AVX-512, the performance difference could be anywhere from insignificant to 20% average speedup depending on the type of algorithm and the additional computation done per individual query.

maumueller:
I see, AVX-512 isn't supported by the platforms that we use to run the benchmark on, but that might of course change :-)

maumueller:
One additional comment: Could you please update README.md and add your implementation to it?

ShikharJ:
@maumueller Sounds good, that should be done within a few days.

ShikharJ:
@maumueller Sorry to disturb you again, would it be possible for you to provide me with the access credentials to the S3 bucket for the same?

maumueller:
Hi @ShikharJ! I'll try to figure out a way to share it with you. @erikbern is the owner, so I don't have rights and Erik might have good reasons for not making access public :-) Let's take it offline via email.

maumueller:
and PS: I was only refering adding your implementation to the list of implementations, not updating the plots at the moment.

ShikharJ:
Ah, I see. I'll do that, thanks.

ShikharJ:
@maumueller The changes are done now. Thanks for the help. I'd be grateful if you could re-run the Travis on the latest commit, since I've added a DiskANN configuration to the build.

erikbern:
Nice – is this ready to be merged? I'm not sure why the CI is failing but it seems good to me

ShikharJ:
@erikbern Yes, no further changes from my side.

erikbern:
Thanks!

PR #233, opened 2021-05-07
Migrate from Travis to Github Actions

GuilhemN:
Hi,

Since TravisCi for Organizations will shut down in the next few weeks, I thought it would be better to move to Github Actions.

There are a few failures, but I suppose that's similar as when using TravisCI.

What do you think?

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,92 @@
+name: ANN benchmarks
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - library: annoy
+            dataset: random-xs-20-angular
+          - library: dolphinn
+            dataset: random-xs-20-angular
+          - library: faiss
+            dataset: random-xs-20-angular
+          - library: flann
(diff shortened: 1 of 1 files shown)


erikbern:
Yeah that sounds like a good idea! I was actually just using Github Actions for the first time last week to set up some stuff so it makes sense. Happy to give it a shot.

GuilhemN:
I think it needs to appear on the default branch to be detected by github and work for later pull requests.

You can check the result the result on my fork meanwhile if you'd like: https://github.com/GuilhemN/ann-benchmarks/actions/runs/819507151

A small note however, Github recently introduced a change that forces to manually trigger CI for first time contributors,  they made an article on that: https://github.blog/changelog/2021-04-22-github-actions-maintainers-must-approve-first-time-contributor-workflow-runs/

erikbern:
let's try merging it and see if it works :)

erikbern:
Looks like it's running: https://github.com/erikbern/ann-benchmarks/actions/runs/840403509

Do you mind updating the README so it points to the Github actions url instead of Travis?

Thanks for doing this!!!!

GuilhemN:
There you go https://github.com/erikbern/ann-benchmarks/pull/236 :smiley:

PR #234, opened 2021-05-09
Fix puffinn algorithm for jaccard similarity measure

GuilhemN:
It seems the dimensions passed to PUFFINN was broken when using Jaccard similarity measure (I couldn't execute it on kosarak).

Here is some partial results (executed on my computer for now) for the Jaccard similarity on the Kosarak dataset.
plot

Would you be interested in having other datasets for jaccard similarity ?

I also believe it would be interesting to differenciate the way datasets are stored for jaccard vs other measures as currently it is quite (no sparse representation).

### DIFF
--- a/ann_benchmarks/algorithms/puffinn.py
+++ b/ann_benchmarks/algorithms/puffinn.py
@@ -15,12 +15,19 @@ class Puffinn(BaseANN):
         self.hash_args = hash_args
 
     def fit(self, X):
+        if self.metric == 'angular':
+            dimensions = len(X[0])
+        else:
+            dimensions = 0
+            for x in X:
+                dimensions = max(dimensions, max(x)+1)
+
         if self.hash_args:
-            self.index = puffinn.Index(self.metric, len(X[0]), self.space,\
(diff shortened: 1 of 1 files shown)


maumueller:
Thanks!

So far there has not been much interest in that part of ann-benchmarks, I'm happy that you picked it up! I still have preprocessing code for all of the datasets in http://www.vldb.org/pvldb/vol9/p636-mann.pdf somewhere in a branch, but it would be great if you could add some others.

With respect to the dataset format: I couldn't figure out a better way to keep the hdf5 format, but there probably is one. (Coming to mind are just two arrays, one with the elements, one with the (start, end) of the individual sets.) The dense format is overkill and, as you might have seen, the first thing ann-benchmarks does is to transform it to a sparse format.

GuilhemN:
Thanks for the advise! I implemented your idea in https://github.com/erikbern/ann-benchmarks/pull/235, it should be easier to add new sparse datasets with this format.

maumueller:
Thanks!

PR #235, opened 2021-05-12
Use sparse datasets for jaccard similarity

GuilhemN:
Following @maumueller's comment in https://github.com/erikbern/ann-benchmarks/pull/234#issuecomment-836243852, here is an implementation of sparse datasets to be closer to jaccard similarity requirements.

Note however that this change is not backward compatible for the kosarak dataset.
Is it a no go and should I add some backward compatibility layers?

### DIFF
--- a/ann_benchmarks/datasets.py
+++ b/ann_benchmarks/datasets.py
@@ -2,7 +2,6 @@ import h5py
 import numpy
 import os
 import random
-import sys
 
 from urllib.request import urlopen
 from urllib.request import urlretrieve
@@ -34,18 +33,24 @@ def get_dataset(which):
             print("Creating dataset locally")
             DATASETS[which](hdf5_fn)
     hdf5_f = h5py.File(hdf5_fn, 'r')
-    return hdf5_f
+
+    dimension = hdf5_f.attrs['dimension'] if 'dimension' in hdf5_f.attrs else len(hdf5_f['train'][0])
+
+    return hdf5_f, dimension
 
 
 # Everything below this line is related to creating datasets
 # You probably never need to do this at home,
 # just rely on the prepared datasets at http://ann-benchmarks.com
 
+
 def write_output(train, test, fn, distance, point_type='float', count=100):
     from ann_benchmarks.algorithms.bruteforce import BruteForceBLAS
     n = 0
     f = h5py.File(fn, 'w')
+    f.attrs['type'] = 'dense'
     f.attrs['distance'] = distance
+    f.attrs['dimension'] = len(train[0])
     f.attrs['point_type'] = point_type
     print('train size: %9d * %4d' % train.shape)
     print('test size:  %9d * %4d' % test.shape)
@@ -56,10 +61,8 @@ def write_output(train, test, fn, distance, point_type='float', count=100):
     neighbors = f.create_dataset('neighbors', (len(test), count), dtype='i')
     distances = f.create_dataset('distances', (len(test), count), dtype='f')
     bf = BruteForceBLAS(distance, precision=train.dtype)
-    train = dataset_transform[distance](train)
-    test = dataset_transform[distance](test)
+
     bf.fit(train)
-    queries = []
     for i, x in enumerate(test):
         if i % 1000 == 0:
             print('%d/%d...' % (i, len(test)))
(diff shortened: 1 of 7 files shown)


maumueller:
Thanks, no backward compatibility is needed, it wasn't a good decision back then. I'll comment on the PR later :-)

ann_benchmarks/datasets.py
```diff
@@ -34,18 +33,24 @@ def get_dataset(which):
             print("Creating dataset locally")
             DATASETS[which](hdf5_fn)
     hdf5_f = h5py.File(hdf5_fn, 'r')
-    return hdf5_f
+
+    dimension = hdf5_f.attrs['dimension'] if 'dimension' in hdf5_f.attrs else len(hdf5_f['train'][0])
```
maumueller, line 37: I would like to see a comment here that this is for reasons of backwards compatibility.

ann_benchmarks/datasets.py
```diff
@@ -69,10 +72,49 @@ def write_output(train, test, fn, distance, point_type='float', count=100):
...
+    f = h5py.File(fn, 'w')
+    f.attrs['type'] = 'sparse'
+    f.attrs['distance'] = distance
+    f.attrs['dimension'] = dimension
+    f.attrs['point_type'] = 'bit'
+    print('train size: %9d * %4d' % (train.shape[0], dimension))
+    print('test size:  %9d * %4d' % (test.shape[0], dimension))
+
+    train = numpy.array(list(map(sorted, train)))
+    test = numpy.array(list(map(sorted, test)))
```
maumueller, line 89: I think we need to add somewhere in the documentation that elements in sets are guaranteed to be ordered. I think this can speed up some filtering approaches.

ann_benchmarks/distance.py
```diff
@@ -48,9 +48,17 @@ def transform_dense_to_sparse(X):
...
+        X.append(data[index:index+l])
+        index += l
+    
+    return X
+
+def dataset_transform(dataset):
+    if dataset.attrs.get('type', 'dense') != 'sparse':
+        return np.array(dataset['train']), np.array(dataset['test'])
+    
+    return sparse_to_lists(dataset['train'], dataset['size_train']), sparse_to_lists(dataset['test'], dataset['size_test'])
```
maumueller, line 64: We might want to add a comment about the sparse encoding that is used here.

ann_benchmarks/datasets.py
```diff
@@ -311,41 +353,29 @@ def kosarak(out_fn):
...
-                X[i][ids[int(x)]] = 1
-            i += 1
-
-    X_train, X_test = train_test_split(numpy.array(X), test_size=500)
-    write_output(X_train, X_test, out_fn, 'jaccard', 'bit')
+                X.append(list(map(int, line.split())))
+                dimension = max(dimension, max(X[-1]) + 1)
+
+    X_train, X_test = train_test_split(numpy.array(X), test_size=500, dimension=dimension)
+    write_sparse_output(X_train, X_test, out_fn, 'jaccard', dimension)
```
maumueller, line 368: Note that the original code only included sets that include at least 20 elements to make sure that queries are somewhat interesting.

maumueller (approved):
Very nice work, @GuilhemN. This makes much more sense than the original dense encoding. Nice minimally-invasive changes  👍 Do you think it would make more sense to have sets instead of lists as the standard data type for jaccard?

maumueller:
Please also not that CI is failing because of plotting problems with the new code <https://travis-ci.org/github/erikbern/ann-benchmarks/jobs/770897666#L2079-L2084>

ann_benchmarks/datasets.py
GuilhemN: The diff is not very explicit, but it's still here, see https://github.com/erikbern/ann-benchmarks/pull/235/files/72019680f36a2b80856bdc3fd7e6584c49930b0b#diff-6a33a2ec7e432a5d84bba3fee0f59ec57fc44f3786a8fe1c75b6a3f617270ee2R363

ann_benchmarks/datasets.py
GuilhemN: I added a note in the README in replacement of the former one about no longer supporting set similarity. Is this what you had in mind?

GuilhemN (commented):
Thank you for your comments!

> Do you think it would make more sense to have sets instead of lists as the standard data type for jaccard?

I'm not sure, for the algorithms currently supported, lists are expected by puffinn, for datasketch we just need to be able to iterate over the data, and pynndescent expects a scipy sparse matrix so we would still need to convert the data.
So I'd say maybe later if other algorithms use sets but for now it seems unnecessary.

(And I don't know very well the interface C <-> python, but wouldn't it make the communication harder for algorithms in C?)

ann_benchmarks/datasets.py
maumueller: I think that's good, yes.

ann_benchmarks/datasets.py
maumueller: Ah, sorry. I missed that part in the diff :-)

maumueller:
Thanks so much @GuilhemN. Do you have any particular datasets in mind that you would like to add?

GuilhemN:
I'm doing an internship at EPFL that aims at benchmarking different approaches for KNN construction with Jaccard similarity, and I'm particularly interested in the following paper https://hal.inria.fr/hal-02357950/file/Fingerprinting%20Big%20Data%3A%20The%20Case%20of%20KNN%20Graph%20Construction.pdf.

So I think I'll implement the datasets they used first (Movielens, AmazonMovies, DBLP, Gowalla) but I'm opened to other suggestions if you have any :)

maumueller:
I have MovieLens and DBLP lying around somewhere from another project. I wouldn't add too many datasets for now since Jaccard isn't the main focus and I'd first like to see more implementations supporting it.

With regard to the paper [outside of scope for ann-benchmarks]: It's funny, they just seem to be doing feature hashing, the oldest trick in ML. Also, their minhash implementation seems to be actually using permutations, which is a really bad idea for such sparse data.

GuilhemN:
> I have MovieLens and DBLP lying around somewhere from another project. I wouldn't add too many datasets for now since Jaccard isn't the main focus and I'd first like to see more implementations supporting it.

Movielens support is on its way :)
I've also seen that NMSLIB supports Jaccard similarity so I'm currently working on making it work with ann-benchmarks. But globally I haven't seen that many libraries supporting jaccard similarity.

> With regard to the paper [outside of scope for ann-benchmarks]: It's funny, they just seem to be doing feature hashing, the oldest trick in ML. Also, their minhash implementation seems to be actually using permutations, which is a really bad idea for such sparse data.

Indeed, the idea is quite simple. It seems to work quite well from their experiments, but as you said the implementations of some alternatives are not so good so I thought I would try to implement their idea it in some real world libraries to check whether it's actually efficient. (My internship is quite short and I don't know so well the field yet so it seemed both more efficient and objective).

So far I compared it against http://ieee-focs.org/FOCS-2017-Papers/3464a663.pdf in a repository of my own and it seemed to compete quite well in terms of speed/relative error, but the implementation is mine as I couldn't find one from the author.


Thanks for your comments!

PR #237, opened 2021-05-18
Remove Travis

erikbern:
Now thanks to #233 we don't need it anymore


erikbern:
@GuilhemN do we need both push and pull_request as events? Seems like it creates 2 jobs for everything. Maybe only the latter is enough?

GuilhemN:
If you want to use branches on this repository, you can limit the branches that trigger events (like [here](https://github.com/nelmio/NelmioApiDocBundle/blob/master/.github/workflows/continuous-integration.yml#L6-L14) for instance), would that do the job?

I believe we need both at least on master as otherwise it wouldn't have a clear build status.

erikbern:
> I believe we need both at least on master as otherwise it wouldn't have a clear build status.

If we never push to master directly then maybe it's enough to just use pull_request?

erikbern:
Merging this for now, we should fix the failing builds separately

PR #238, opened 2021-05-18
Make Puffinn algorithm compatible with sparse dataset format

GuilhemN:
Sorry my two PRs (https://github.com/erikbern/ann-benchmarks/pull/235 and https://github.com/erikbern/ann-benchmarks/pull/234) were not compatible with each other, I should have added a note.
So here's a fix, and I added puffinn with jaccard to the tests set to make it easier to detect in the future.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -59,6 +59,8 @@ jobs:
             dataset: random-xs-20-angular
           - library: puffinn
             dataset: random-s-jaccard
+          - library: pynndescent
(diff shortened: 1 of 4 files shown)


maumueller:
Thanks, and thanks for adding it to CI. Is pynndescent affected in a similar way?

GuilhemN:
Pynndescent is working fine, it was already compatible with numpy arrays.
But I added it to CI as well, it is safer to test it automatically.

maumueller:
Great, thanks!

PR #239, opened 2021-05-19
Add movielens support and Jaccard similarity with NMSLIB

GuilhemN:
NMSLIB supports jaccard similarity so I thought it would be great to support it.

I plotted the results for single queries on movielens1m:
plot

I'll probably slow down the rhythm after this PR as my modifications won't be as generic and rather applied to my problem.


GuilhemN:
By the way I'm used to break my PRs into small parts to ease reviewing but do not hesitate if you'd rather have bigger PRs and less often / or just on specific subjects.

ann_benchmarks/datasets.py
```diff
@@ -424,6 +424,60 @@ def lastfm(out_fn, n_dimensions, test_size=50000):
...
+def movielens20m(out_fn):
+    import zipfile
+
+    url = 'http://files.grouplens.org/datasets/movielens/ml-20m.zip'
+    fn = 'ml-20m.zip'
+    download(url, fn)
+    with zipfile.ZipFile(fn) as z:
+        f = z.open('ml-20m/ratings.csv')
+        f.readline()# ignore the header
+        movielens(f, out_fn, ',')
```
maumueller, line 480: Couldn't that be done by a single method by passing the file names as parameters? It looks somehow suspicious that ml-20m is different to 10m and 1m?

algos.yaml
```diff
@@ -1014,3 +1014,74 @@ bit:
...
+          arg-groups:
+                - {"tuneK": 10, "desiredRecall": [0.99, 0.97, 0.95, 0.9, 0.85, 0.8,
+                    0.7, 0.6, 0.5, 0.4, 0.3, 0.2, 0.1]}
+                - False
+          # This run group produces thirteen algorithm instances:
+          # NmslibNewIndex("angular", "vptree", {"tuneK": 10,
+          # "desiredRecall": 0.99}), NmslibNewIndex("angular", "vptree",
+          # {"tuneK": 10, "desiredRecall": 0.97}), and so on up to
+          # NmslibNewIndex("angular", "vptree", {"tuneK": 10, "desiredRecall":
+          # 0.1}).
```
maumueller, line 1087: @searchivarius: Sorry for cross-including you like this, but do you think these parameter settings make sense?

ann_benchmarks/datasets.py
GuilhemN: Well I'm not sure of the why, but ml-20m is a csv, while 10m and 1m are .dat files, so the format is a bit different and I tried to merge their processing but it seems complicated to do it further (unless I pass an option to remove the headers of the csv file ?)

maumueller: Precisely, what about `def movielens(out_fn, fn, ignore_header=False)`?

GuilhemN: Since the zip files do not use the same hierarchy I added a parameter `ratings_file`, and for the different separator I added `separator`, giving:
```
movielens(fn, ratings_file, out_fn, separator='::', ignore_header=False)
```

would that be acceptable ?

erikbern:
nice – happy to merge whenever

maumueller:
Merging this created many conflicts with other work done on nmslib's jaccard integration. I'm closing this and cherry-pick the movielens changes to #304.

PR #243, opened 2021-05-21
Revert "Update NGT parameters"

masajiro:


masajiro:
The images in this were unintentionally updated by our results. However, these results were exactly produced on a c5.4xlarge where the timeout is 5 hours.

PR #244, opened 2021-05-22
Update parameters for ONNG and QG

masajiro:
Sorry for the confusion. This PR is the same as #221 except for our results.


erikbern:
thanks!

PR #245, opened 2021-05-23
try to fix sptag

erikbern:
I'm on an M1 mac and I realized almost nothing works because of weird Docker issues, so I'm not able to test it. Hopefully this works.


erikbern:
Seems like it did the trick. Only opendistroknn, diskann, and mih left now.

PR #248, opened 2021-06-09
Add Vespa HNSW to ANN Benchmarks

toregge:
[Vespa](https://vespa.ai/) is an open source big data serving engine that supports approximate nearest neighbor search by using an in-memory HNSW index. Vespa implements a modified version of the Hierarchical Navigable Small World (HNSW) graphs algorithm ([paper](https://arxiv.org/abs/1603.09320)). The Vespa implementation supports:

- Combining HNSW lookups with metadata filters and text search
- Real-time index updates
- Multi-threaded indexing and searching

For more details see [Approximate Nearest Neighbor Search using HNSW Index](https://docs.vespa.ai/en/approximate-nn-hnsw.html).

This integration adds benchmarking directly towards the Vespa HNSW index and experiments show comparable performance with other HNSW algorithms. The Vespa HNSW index has less co-localisation of data in memory due to the extra functionality supported, and the performance differences with the other HNSW algorithms are mainly due to cache misses. The experiments were conducted on an Ubuntu 18 VM with Intel i7-8700k CPU and 48 GB of memory.

glove-100-angular (k = 10):
image

sift-128-euclidean (k = 10):
image

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -61,6 +61,8 @@ jobs:
             dataset: random-s-jaccard
           - library: pynndescent
             dataset: random-s-jaccard
+          - library: vespa
+            dataset: random-xs-20-angular
       fail-fast: false
 
     steps:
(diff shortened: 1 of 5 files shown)


erikbern:
Very cool! Can you add it to https://github.com/erikbern/ann-benchmarks/blob/master/.github/workflows/benchmarks.yml so it runs in the CI suite as well?

toregge:
> Very cool! Can you add it to https://github.com/erikbern/ann-benchmarks/blob/master/.github/workflows/benchmarks.yml so it runs in the CI suite as well?

Done.

erikbern:
Thanks!

jobergum:
Thanks a lot @erikbern

PR #258, opened 2021-08-25
Add scipy ckdtree to the benchmark

VincentRouvreau:


VincentRouvreau:
> n high-dimensional A

@maumueller the recall will always be 1, but no real idea if it is really efficient (especially in high dimensions). I tried to build the figures, but it takes too much time...

PR #264, opened 2021-10-13
updating the change in Docker's python API

alex--m:
See [Docker SDK for Python](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.wait)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 ansicolors==1.1.8
-docker==2.6.1
+docker==5.0.2
 h5py==2.7.1
 matplotlib==2.1.0
 numpy==1.13.3


alex--m:
> Thanks for the contribution! Looking at CI (e.g., https://github.com/erikbern/ann-benchmarks/pull/264/checks?check_run_id=3882613149#step:5:3664) this doesn't seem to work well in the environment used there.
> 
> Doesn't this change imply that the docker module should be updated at https://github.com/erikbern/ann-benchmarks/blob/master/requirements.txt#L2?

You're absolutely right... when I installed from `requirements.txt` it put docker v2.6.1 alongside what I have, which is v5.0.2, and then when I ran it without venv I got the error I fixed in this commit... Does it make sense to upgrade the requirements file to the latest docker SDK version?

erikbern:
merging this for now, the comment i left isn't super important

Issue #280, opened 2021-11-30
Building image for faiss algorithm fails


carlryn:
`python install.py --algorithm faiss` fails with the error:

> ERROR: Cannot uninstall 'PyYAML'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
> The command '/bin/sh -c python3 -m pip install -r requirements_py38.txt' returned a non-zero code: 1

When installing the requirements at line: [RUN python3 -m pip install -r requirements_py38.txt](https://github.com/erikbern/ann-benchmarks/blob/9966b9020eacd432b89f41a13f0c5d82dd6870e4/install/Dockerfile.faiss#L12)

It fails when building the faiss image. Tried also building it manually without install.py with same error. Assuming, this is due to the switch to conda. Probably has something to do with the following commit: 
https://github.com/erikbern/ann-benchmarks/commit/dfa2abc81bcc868013bcfce8e434a225f9ea709a#diff-d6599cc06b437f681f7fbce3207c0f5d9347cfea4d5b52ea81bb40e55fccdec0

Mixing conda and pip is tough :P

carlryn:
The following definition will at least run.
```
FROM ann-benchmarks

RUN apt update && apt install -y wget
RUN wget https://repo.anaconda.com/archive/Anaconda3-2020.11-Linux-x86_64.sh
RUN bash Anaconda3-2020.11-Linux-x86_64.sh -b

ENV PATH /root/anaconda3/bin:$PATH
RUN conda install -c pytorch faiss-cpu
COPY requirements_py38.txt ./
RUN conda install -c anaconda pandas
# conda doesn't like some of our packages, use pip
# https://developpaper.com/a-pit-of-mkl-library-in-linux-anaconda/
ENV LD_PRELOAD /root/anaconda3/lib/libmkl_core.so:/root/anaconda3/lib/libmkl_sequential.so 

RUN python3 -c 'import faiss; print(faiss.IndexFlatL2)'
```

maumueller:
Thanks @carlryn (👋 fellow CPHer)

This happens because conda installs pyyaml==5.3.1 and we try to raise it to 5.4 a couple of commits ago. I'l take a look at it in #281

carlryn:
I found https://github.com/harsha-simhadri/big-ann-benchmarks/blob/main/install/requirements_conda.txt
which at leasts builds the requirements and I can start executing. 
My faiss dockerfile became:
[...]

PR #281, opened 2021-11-30
Update Dockerfile.faiss

maumueller:


maumueller:
Closes #280 

(There are some unrelated CI problems, such as CI running the faiss gpu implementation and mih failing.)

Issue #282, opened 2021-12-02
Container exit code is not an integer, but a dictionary


carlryn:
Hej,

Was executing the faiss-ivf and error after what seems to be a succesfull benchmarking run.

> 2021-12-02 15:45:57,143 - annb.0e88ef45e8 - ERROR - Container.wait for container 0e88ef45e8 failed with exception
> Traceback (most recent call last):
>   File "/home/ubuntu/git/ann-benchmarks/ann_benchmarks/runner.py", line 269, in run_docker
>     logger.error('Child process for container %s raised exception %d' % (container.short_id, exit_code))
> TypeError: %d format: a number is required, not dict

I changed the line providing the error message to:
`logger.error(f'Child process for container {container.short_id} raised exception {exit_code}')`

This gives the following output:

> 2021-12-02 15:54:36,931 - annb.a277890f6b - ERROR - Child process for container a277890f6b raised exception {'Error': None, 'StatusCode': 0}

Now, changing the line https://github.com/erikbern/ann-benchmarks/blob/b95dda65146156ee93c52e0265c8ee5b62bd11dd/ann_benchmarks/runner.py#L265
to: `if exit_code["StatusCode"] != 0:` fixes it.

carlryn:
nvm I can see a pull request that actually fixes this...

carlryn:
Actually the pull request should be accepted.

carlryn:
And how do I open this, it should not be closed until the pull request is merged in

erikbern:
I think it's a problem where different versions of Docker returns different things. If you fix it, can you handle both the dict case and the integer case? Thanks!

carlryn:
I'll fix it.

jgoed:
Hi everyone, 
I'm confused... If this issue is fixed, how does it come that I encounter the exact same problem today using the "main" branch? Note that manually fixing the line in runner.py as purposed above still fixes it.

PR #283, opened 2021-12-03
Handle docker container return values

carlryn:
Hej,

This fixes issue https://github.com/erikbern/ann-benchmarks/issues/282
I've unit-tested the suggested method `_handle_container_return_value` only locally since it seems that you don't tests :P 

i also ran the following: `python run.py --algorithm faiss-ivf --runs 1 --dataset random-xs-20-angular --run-disabled --timeout 300`
and it executes beautifully. 

Please let me know if you want some changes.


erikbern:
thanks!

PR #286, opened 2021-12-30
Ubuntu 20.04 and default python

erikbern:
In an attempt to simplify the installation, let's just use latest ubuntu (20.04) with the default version of Python (3.8)


erikbern:
Merging this – remaining problems have to do with individual algorithms

PR #289, opened 2022-01-21
Sparse jaccard update

lmcinnes:
This is an update following some of the changes to the sparse jaccard datasets and benchmarking with kosarak-jaccard. I have tried to support both the format the data currently flows through the algorithms in, and the intended format based on my interpretation of #235 in terms of the intended format (a kind of variation of CSR format essentially; passed in row by row). I also added nmslib support for sparse jaccard, so now there should be at least 6 algorithms/implementations running on the sparse jaccard data. NGT does support sparse jaccard at some level, but not at a stage that can be integrated with this benchmark. Hopefully in the future we can add NGT support for sparse jaccard as well.

### DIFF
--- a/algos.yaml
+++ b/algos.yaml
@@ -1138,29 +1138,24 @@ bit:
       run-groups:
         NN-40:
           arg-groups:
-            - {"n_neighbors": [40], "diversify_prob": [0.5, 1.0],
-               "pruning_degree_multiplier":[1.5], leaf_size: [120, 240]}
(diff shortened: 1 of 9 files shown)


erikbern:
Yeah I saw that too, not sure what's going on

lmcinnes:
Everything but MIH is working now. Is this mergable? It adds several algorithms (from nmslib) to the jaccard benchmark and supports various formats for the jaccard datasets.

erikbern:
Thanks! I can re-run benchmarks shortly

PR #297, opened 2022-03-09
added --json flag to data_export.py

DvirDukhan:
this PR introduces the ability to export results into a json file

### DIFF
--- a/data_export.py
+++ b/data_export.py
@@ -1,9 +1,21 @@
 import argparse
 import csv
+import json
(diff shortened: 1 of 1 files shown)


maumueller:
Nice work @DvirDukhan. I'm unable to edit the PR, but I suggest that we streamline the command-line arguments. What do you think about removing `--json` and figuring out whether csv/json should be used from the provided filename?

PR #304, opened 2022-07-11
Small bugfixes for jaccard datasets, include MovieLens from #239. 

maumueller:
Looking at #239 I noticed many small bugs when working with Jaccard similarity datasets. This PR fixes those.


maumueller:
I got the following results: 

Kosarak:

kosarak-jaccard

Movielens1M: 

movielens1m-jaccard

Movielens10M:

movielens10m-jaccard

Something seems to be wrong with PyNNDescent. @lmcinnes  I'm not sure what has happened with PyNNDescent? I didn't change the sparse-dense transformation, which seems to be the same as the one used by nmslib.

lmcinnes:
Thanks for the heads up; something has definitely gone astray with pynndescent. I'll see if I can figure it out soon when I get a little time to look into it.

lmcinnes:
I did a fresh checkout and ran things locally (getting the docker images working on the system I had available was more trouble than I wanted). I got the following:
image
So either it is in the actual docker instance that things are going astray, or there is something else wrong. Do you have any of the logs from the run that might give some indication of what went astray? If not I'll try to get access to a more docker friendly system and try the whole process.

maumueller:
Thanks for the fast reply, @lmcinnes. Indeed, it seems that I was using an old docker image. Rebuilding resulted in 
kosarak-jaccard

erikbern:
Impressive job by pynndescent!

PR #306, opened 2022-07-17
Elastiknn: update to latest release, remove data scaling

alexklibisz:
Bumps Elastiknn versions to the latest.

Modifies algo parameters such that we can remove a data scaling step which was previously required in order to get decent results. The main difference is the width param, w: we set it higher to accommodate for the scale of the vector values, instead of scaling the vectors to accommodate for the w param.

New results should be on-par or better for Fashion MNIST. I haven't tested thoroughly with SIFT yet. I'll run some more tests this week on my own hardware and follow-up with another PR if necessary. Even if the params don't work well for SIFT, the algo will early-stop if throughput is < 50 qps after 500 queries.

New results running on my Dell XPS w/ i7-8750H CPU @ 2.20GHz. 

plot

### DIFF
--- a/algos.yaml
+++ b/algos.yaml
@@ -570,14 +570,14 @@ float:
       constructor: L2Lsh
       base-args: []
       run-groups:
-        elastiknn-lsh:
+        elastiknn-l2lsh:
           args:
-          - [50, 75, 100]
-          - [3, 4]
-          - [1, 3, 7]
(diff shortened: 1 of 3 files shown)


maumueller:
Thanks!

erikbern:
nice! maybe time to rerun all benchmarks soon

PR #312, opened 2022-09-02
Switch to CentOS Stream 8 as OS for vespa container.

toregge:
With [Vespa 8](https://blog.vespa.ai/vespa-8-is-here/), the [supported OS has changed from CentOS 7 to CentOS Stream 8](https://blog.vespa.ai/Upcoming-changes-in-OS-support-for-Vespa/).


maumueller:
Thanks!

PR #314, opened 2022-10-29
Introduce custom exception type to recover from query errors

alexklibisz:
This adds a way to let an algo early-stop a set of query parameters that are not going well (or
The query method can throw a BaseANNQueryException, and the `run` method will move onto the next set of query params (instead of killing the entire run).

### DIFF
--- a/ann_benchmarks/algorithms/base.py
+++ b/ann_benchmarks/algorithms/base.py
@@ -17,6 +17,10 @@ class BaseANN(object):
         pass
 
     def query(self, q, n):
+        """
+        Run a a single query and return the array of candidate indices.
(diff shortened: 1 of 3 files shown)


maumueller:
Thanks Alex, that's a nice idea. I do wonder if (in the long run) we shouldn't also write a result file for these configurations. At the moment, it will fail but it will these configurations again, which should be avoided.

All CI runs fail, not sure why.

alexklibisz:
Yeah, still thinking through whether this makes sense. Feels a bit hacky. Will look into CI. No rush to merge this.

alexklibisz:
Closing for now. Planning to revisit this weekend and will re-open if this approach still makes sense.

PR #315, opened 2022-10-31
Add Lucene 9.4.1 Knn algorithm runner and docker file

benwtrent:
This commit adds support for running Lucene's KNN search. The trick here is around Lucene requiring the JVM and allowing testing through Python. [PyLucene](https://lucene.apache.org/pylucene/) python library is utilized for interacting between Python and the JVM.

Building and using PyLucene can be laborious. There are well documented steps on the PyLucene site. The Docker image takes care of all the heavy lifting there.


ann_benchmarks/algorithms/luceneknn.py
```diff
@@ -0,0 +1,120 @@
...
+        if self.metric == 'angular':
+            X = sklearn.preprocessing.normalize(X, axis=1, norm='l2')
+        self.queries = X.tolist()
+        self.topK = n
+
+    def run_batch_query(self):
+        self.res = []
+        append = self.res.append
+        n = self.topK
+        num_candidates = self.topK + self.fanout
```
jtibshirani, line 102: In the Lucene benchmarks, the primary parameter is "fanout" and we use `k + fanout` candidates. But usually the search-time parameter is called "efConst" and it represents the total number of candidates used in search.

To be consistent with the other libraries like hnswlib, maybe we could rename `self.fanout` to `self.ef_const` and use it directly as the number of candidates.

jtibshirani (commented):
It's great you found a way to minimize the overhead!

Am I understanding right that this runs the Lucene 9.1 version (whereas current Lucene version is 9.4)? After 9.1 we found a difference in the algorithm that made the latency/ recall curve quite different from other HNSW implementations: https://issues.apache.org/jira/browse/LUCENE-10527. I wonder if we could prepare a more recent version so the results are more up-to-date?

msokolov:
PyLucene sounds like a better approach to this for sure, although we might need to get a release of it out? Release cadence there seems to be roughly every six months, so it seems due?

I would still be somewhat concerned about warming up the JVM so the HotSpot compiler can work its magic. It sounds as if you got decent results, but I wonder how you are able to do warm-up as the ann-benchmarks framework really has no provision made for that.

maumueller:
Thanks for the contribution! Is this ready to be merged? 

> > It sounds as if you got decent results, but I wonder how you are able to do warm-up as the ann-benchmarks framework really has no provision made for that.

> For sure, we could do this in the "fit" phase so that warmup doesn't change search times. I was mostly concerned about moving forward first and treating additional things as optimizations.

I'm not sure of the details in the lucene implementation (looks like hnsw from the parameter settings?), but most indexes will use their query method for building the index. So ,building/fitting the dataset should do the warm-up. If it's indeed a concern, we provide the `--runs` argument that will repeat querying for all query points (the part starting here: https://github.com/erikbern/ann-benchmarks/blob/master/ann_benchmarks/runner.py#L28)

benwtrent:
@maumueller It is not ready for merge yet. Thank you for checking!

We are seeing about updating the Lucene implementation tested. That way we don't have to do an update PR immediately after :)

erikbern:
Happy to merge whenever this is ready!

ann_benchmarks/algorithms/luceneknn.py
```diff
@@ -0,0 +1,124 @@
...
+        self.searcher = IndexSearcher(DirectoryReader.open(self.dir))
+
+    def set_query_arguments(self, ef_const):
+        self.name = f"luceneknn-{self.dimension}-{self.param['M']}-{self.param['efConstruction']}-{ef_const}"
+        self.ef_const = ef_const
+
+    def query(self, q, n):
+        if self.metric == 'angular':
+            q = q / np.linalg.norm(q)
+        return self.run_knn_query(n + self.ef_const, n, q.tolist())
```
jtibshirani, line 94: This should just be `self.ef_const` instead of `n + self.ef_const`. (It's confusing because in the Lucene tests we use a different concept called "fanout".)

ann_benchmarks/algorithms/luceneknn.py
```diff
@@ -0,0 +1,124 @@
...
+        for q in self.queries:
+            append(self.run_knn_query(num_candidates=num_candidates, n=n, q=q))
+
+    def get_batch_results(self):
+        return self.res
+
+    def run_knn_query(self, num_candidates, n, q):
+        query = KnnVectorQuery("knn", JArray('float')(q), num_candidates)
+        topdocs = self.searcher.search(query, n)
+        return [d.doc for d in topdocs.scoreDocs]
```
jtibshirani, line 116: I just noticed -- this seems fragile because doc IDs aren't guaranteed to match the insertion order. For example, two segments could get merged together, which could totally reorder the document IDs. I noticed that you added an `id` stored field, maybe you intend to be using it here?

benwtrent: 🤦

ann_benchmarks/algorithms/luceneknn.py
```diff
@@ -0,0 +1,124 @@
...
+            q = q / np.linalg.norm(q)
+        return self.run_knn_query(num_candidates=self.ef, n=n, q=q.tolist())
+
+    def prepare_batch_query(self, X, n):
+        if self.metric == 'angular':
+            X = sklearn.preprocessing.normalize(X, axis=1, norm='l2')
+        self.queries = X.tolist()
+        self.topK = n
+
+    def run_batch_query(self):
```
jtibshirani, line 102: What's the difference between `run_batch_query` and `batch_query`?

ann_benchmarks/algorithms/luceneknn.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Custom codec so that the appropriate Lucene94 codec can be returned with the configured M and efConstruction
+    """
+    def __init__(self, M, efConstruction):
+        super(Codec, self).__init__()
+        self.M = M
+        self.efConstruction = efConstruction
+
+    def getKnnVectorsFormatForField(self):
+        Lucene94HnswVectorsFormat(self.M, self.efConstruction)
```
jtibshirani, line 32: I think we need to explicitly `return` here, otherwise there's no return value and Lucene will always use the default parameters for M and efConstruction. (Maybe this is why we were seeing unexpected recall values in our tests?)

jtibshirani (commented):
Hopefully last round of comments!

ann_benchmarks/algorithms/luceneknn.py
benwtrent: Literal LOL!

ann_benchmarks/algorithms/luceneknn.py
benwtrent: See: https://github.com/erikbern/ann-benchmarks/blob/6ad14dc08ea2fa572df7339a9f024975bac1a309/ann_benchmarks/runner.py#L54

One runs the preparation, the other doesn't. Honestly, I don't know when the runner chooses one path or the other, but satisfied both as it isn't difficult.

jtibshirani: hmm, it's just a bit confusing since one implementation is parallel and the other is serial.

jtibshirani (approved):
Looks good to me! You probably want to exclude the `mnist` image though, probably committed by accident.

I haven't had the chance to pull it down and try it yet, but I'll do that soon.

ann_benchmarks/algorithms/luceneknn.py
benwtrent: Actually, this isn't the only problem. I misunderstood PyLucene. Some more work is required. I need to write a pylucene Python extension point and hook into that...

benwtrent:
@erikbern Whenever you are ready, this is ready to merge :)

:shipit:

erikbern:
Looks like the build step is failing – any idea why?

```
Step 13/24 : RUN python3 ./setup.py build
 ---> Running in 0775a35198bb
  File "./setup.py", line 100, in <module>
    ''' %(JDK[platform]))
RuntimeError: 

Java JDK directory '/usr/lib/jvm/temurin-17-jdk-arm64' does not exist.

Please set the environment variable JCC_JDK to the correct location before
running setup.py.
```

benwtrent:
@erikbern I wonder if this is due to platform differences? I was building the image on an M1 macbook. So, the image was targeting the ARM architecture (note the `arm64` in the JVM path). 

What is the preferred architecture? I can try and fix it.

FWIW, I just built it without cache locally just fine.

benwtrent:
@erikbern just pushed a change, I tested it as follows:
```
docker build --no-cache --platform x86_64 --rm -t ann-benchmarks-luceneknn -f install/Dockerfile.luceneknn .
```

It built successfully. Hopefully it works now :)

erikbern:
Thanks! It's failing on something else now:

```
 Step 14/25 : RUN python3 ./setup.py build
 ---> Running in fd8e6ed085c1
  File "./setup.py", line 100, in <module>
    ''' %(JDK[platform]))
RuntimeError: 

Java JDK directory '/usr/lib/jvm/temurin-17-jdk-' does not exist.
```

benwtrent:
Hey @erikbern ! I am not sure why TARGETARCH is empty for your run. I statically set `amd64` for the JVM kind.

If this still doesn't work, could you tell me the architecture on which these images are built and ran? That way I can just statically set it to the correct one in the file.

erikbern:
Looks like it's working now! THanks!

PR #327, opened 2023-01-30
Improvements to CI workflow

alexklibisz:
* Split the "install dependencies" step to two steps: 1) install project dependencies, 2) build the library image.
* Run `pip install` with `--quiet` to limit logging for faster CI debugging.
* Make install.py return 1 if the installation fails. Without this, the CI failures are misleading: install step succeeds even though install failed, then the run step fails w/ a confusing "nothing to run" error.
* Dedup the env vars in the CI workflow. We can have one env block for the full job, rather than the same env vars per step.
* Add concurrency settings to the CI workflow to cancel the workflow for the previous commit when a new commit is pushed.
* Add a timeout-minutes set to 20 to prevent runaway workflows, like the one here: https://github.com/erikbern/ann-benchmarks/pull/328#issuecomment-1408026679

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -2,10 +2,15 @@ name: ANN benchmarks
 
 on: [push, pull_request]
 
+# Cancel the workflow for the previous commit when the new commit is pushed.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: True
+
 jobs:
   build:
-
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


erikbern (approved):
Nice! Happy to merge this any time, just let me know when it's done

alexklibisz:
All done, go for it!

alexklibisz:
@erikbern or @maumueller Could you go ahead and merge this?

erikbern:
Sure let's merge!

erikbern:
Thanks @alexklibisz!!

Issue #316, opened 2022-11-03
Upgrade Ubuntu and Python


alexklibisz:
The repo is on Ubuntu 18 and Python 3.6. 

I'm having trouble upgrading the Elastiknn benchmarks because the latest Elasticseach Ubuntu installer relies on some things that don't seem to be in Ubuntu 18. I've also jumped through some hoops with Python, e.g., intentionally installing older versions of scipy for compat w/ 3.6.

Any reason not to upgrade? Any dragons to be aware of in case I take a first pass at it?

maumueller:
A long time ago, a few implementations failed on Python 3.8, but I'm not sure of the exact status. The base repo runs Python 3.8 and Ubuntu 20.04, it would be great if the docker images use the same.

erikbern:
Yeah let's upgrade to Python 3.10 (or at least 3.9)

alexklibisz:
I finally took a pass at this in #326, w/ ubuntu 22.04 and python 3.10.

I found that 14/33 fail to build for this combination. I spot checked a few. Some might be fixable w/ changes to the local Dockerfiles. Others seem to require changes to the upstream source (e.g., kgraph). I might be mistaken on the specifics, as this was a pretty quick look. Feel free to double check me.

In any case, what should we do for these failing libraries? Python 3.6 is EOL as of December 2021. Ubuntu 18 is EOL April 2023. I don't really think it's viable for one of us to try to upgrade source in all of these libraries. I see a couple other viable options...

* Just let them fail. If the authors care, they will update.
* Try Ubuntu 20 and Python 3.8. Maybe some more of them will build without serious changes?

LMK what you guys think.

alexklibisz:
Here's a more precise summary of what I'm seeing:

* 6 libraries currently fail on master. [See this action](https://github.com/erikbern/ann-benchmarks/actions/runs/4006679948).
* With ubuntu 20.04, at least 9 libraries fail. It's probably actually 15, though, as [this action](https://github.com/erikbern/ann-benchmarks/actions/runs/4039797834) from [this PR](https://github.com/erikbern/ann-benchmarks/pull/328), still has some libraries running after three hours.
* With ubuntu 22.04, 14 libraries fail. [See this action](https://github.com/erikbern/ann-benchmarks/actions/runs/4039309819).

maumueller:
[...]

PR #328, opened 2023-01-30
Upgrade to Ubuntu 20.04 and Python 3.8

alexklibisz:
Upgrade to Ubuntu 20.04 and Python 3.8, re: #316 

Some of the builds are newly-broken. Some were already broken. More detail in #316.


alexklibisz:
Some of these are still running 3 hours later..

image

alexklibisz:
I also don't see a way to stop the runaway workflow. Even after deleting the branch. Weird.

alexklibisz:
Actually let's maybe wait. I'm testing this out locally with elastiknn and seeing a pretty bad memory leak in the python process, that's causing the container to crash. Maybe it's specific to elastiknn but I need to dig a bit more. The memory leak is definitely in the python process, not the java process, based on just polling `ps` output within the container

erikbern:
That's super strange!

alexklibisz:
Nevermind, I rebased on master and tried again today with Elastiknn and Annoy, and both ran fine. 

I must have messed something up in my local setup. Apologies for the confusion. I cleaned up the commits, so I think we're good to merge here.

erikbern:
Nice! Will merge this assuming most tests are green

erikbern:
A bunch of algorithms are failing with various package/version conflicts right now, but my feeling is that it's not super hard to fix. I'll go through a bunch of them later.

erikbern:
Thanks for this contribution!!

PR #330, opened 2023-02-16
Enable powertools repo when installing vespa-ann-benchmark package.

toregge:
This tracks move of xxhash-libs package to the powertools repo which is disabled by default.


maumueller:
Thanks @toregge! Vespa seems to fail a check, please take a look at https://github.com/erikbern/ann-benchmarks/actions/runs/4193194952/jobs/7269761522 to see what is going on.

maumueller:
Thanks for checking, merging this.

PR #331, opened 2023-02-23
Upgrade elastiknn to 8.6.2.0

alexklibisz:
Upgrading elastiknn to the latest version 8.6.2.0.

I had to change the Elasticsearch installation to just use a tar download instead of using apt and deb files. This let me avoid systemd and sysctl, which I couldn't get working with the latest versions of Elasticsearch (even on newer ubuntu containers).

I still have to do a bit of gymnastics to get the python client installed. This should be resolved by upgrading to newer Python.

I had to increase the time spent waiting for Elasticsearch's to become available. It seems to take longer to initialize on 8.x versions compared to 7.x


alexklibisz:
It seems to be failing in CI. Converting back to a draft while I debug.

erikbern:
thanks@

PR #340, opened 2023-03-09
Update milvus

hhy3:


erikbern:
I think my recent changes broke the Milvus build – let me fix it separately

PR #341, opened 2023-03-10
Use Python 3.9 on CentOS Stream 8.

toregge:
The required python packages cannot be installed when using default python3 on CentOS Stream 8 (Python 3.6.8) but they
can be installed when using Python 3.9.16.


erikbern:
awsome, thanks!

PR #342, opened 2023-03-10
Add Github stargazer badges

loisaidasam:
Leverage https://shields.io/ to create Github stargazer badges next to each evaluated library implementation.

Skip those that are part of bigger projects that don't directly relate to ANN (google-research, scipy, etc.)

While not the most direct metric, it gives some sense of how well-adopted each implementation is.


erikbern:
nice!

PR #347, opened 2023-03-26
Update numpy and scipy to work with Python 3.11

ankane:
Hi, thanks for this project!

This updates NumPy and SciPy to the latest version so `pip install -r requirements.txt` works with Python 3.11.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -82,7 +82,7 @@ jobs:
     - uses: actions/checkout@v2 # Pull the repository
 
     - name: Install OS Dependencies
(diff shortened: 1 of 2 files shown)


ankane:
Also pushed a fix for CI (need to call `apt-get update` before `apt-get install` or else it will occasionally error with `Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?`, like [this](https://github.com/ankane/ann-benchmarks/actions/runs/4525436667/jobs/7969928129)).

maumueller:
Thanks!

ankane:
Thanks @maumueller!

PR #348, opened 2023-03-29
Install dnf-plugins-core to get dnf copr plugin.

toregge:
Docker image for CentOS Stream 8 no longer includes dnf copr plugin.


erikbern:
Thanks!

PR #354, opened 2023-04-07
remove scipy dep

erikbern:
wasn't able to install it on my mac, and it's basically not used anyway – it's very easy to just implement these distance metrics using numpy

### DIFF
--- a/test/test-distance.py
+++ b/test/test-distance.py
@@ -1,0 +1,10 @@
+import unittest
+import numpy
+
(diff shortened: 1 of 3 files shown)


test/test-distance.py
```diff
@@ -0,0 +1,10 @@
+import unittest
```
erikbern, line 1: side note but:
1. let's switch to pytest
2. let's run these as a part of CI

PR #361, opened 2023-04-07
Install sptag from pypi

erikbern:
Building SPTAG is so slow it times out during CI: https://github.com/erikbern/ann-benchmarks/actions/runs/4640595250/jobs/8212702264

I noticed it's on PyPI so let's try that? It's the only algorithm right now that fails in CI and it would be amazing to get the build green again!


erikbern:
this doesn't seem to work :(

PR #365, opened 2023-04-09
pgvector improvements

ankane:
Initial round of improvements for pgvector:

1. Set `ivfflat.probes` to improve recall
2. Remove `OPTFLAGS=""` so it's compiled with `-march=native` (which enables additional SIMD instructions on supported platforms)
3. Use Unix-domain socket (rather than networking stack)
4. Use binary serialization (to reduce serialization overhead)
5. Use prepared statements (to reduce statement parsing overhead)
6. Increase `work_mem` (to allow for more memory usage)

### DIFF
--- a/algos.yaml
+++ b/algos.yaml
@@ -547,6 +547,7 @@ float:
       run-groups:
         pgvector:
           args: [[100, 200, 400, 1000, 2000, 4000, 10000]]
+          query-args: [[1, 2, 4, 10, 20, 40, 100]]
 
   euclidean:
     vamana(diskann):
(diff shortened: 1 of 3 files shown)


erikbern:
Nice! I wasn't able to run it with a unix socket – some permission issue. Did you run into it?

erikbern:
Let me know once this is ready to be merged!

ankane:
Yeah, needed to add the `pg_hba.conf` line to get it working in the Docker image.

It should be in a good spot now to merge.

erikbern:
Thanks! Will rerun in a few days!

PR #366, opened 2023-04-09
Fix warnings in plot.py

ankane:
Fixes a few warnings generated by:

```sh
python3 plot.py --dataset $DATASET --output plot.png --raw
```

Output

```text
ann-benchmarks/plot.py:35: UserWarning: linestyle is redundantly defined by the 'linestyle' keyword argument and the fmt string "-" (-> linestyle='-'). The keyword argument will take precedence.
  (handle,) = plt.plot(
ann-benchmarks/plot.py:40: UserWarning: linestyle is redundantly defined by the 'linestyle' keyword argument and the fmt string "-" (-> linestyle='-'). The keyword argument will take precedence.
  (handle2,) = plt.plot(
ann-benchmarks/plot.py:76: MatplotlibDeprecationWarning: The 'b' parameter of grid() has been renamed 'visible' since Matplotlib 3.5; support for the old name will be dropped two minor releases later.
  plt.grid(b=True, which="major", color="0.65", linestyle="-")
```

### DIFF
--- a/plot.py
+++ b/plot.py
@@ -33,12 +33,12 @@ def create_plot(all_data, raw, x_scale, y_scale, xn, yn, fn_out, linestyles, bat
         max_x = max([max_x] + [x for x in xs if x < 1])
         color, faded, linestyle, marker = linestyles[algo]
         (handle,) = plt.plot(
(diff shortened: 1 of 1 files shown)


erikbern:
Thanks!

PR #369, opened 2023-04-10
Fix KeyError

ankane:
With Docker 23 on Ubuntu 22.04, I'm seeing the following error:

```text
2023-04-10 20:00:28,118 - annb.330fe43cd51c - ERROR - Container.wait for container 330fe43cd51c failed with exception
  File "/tmp/ann-benchmarks/ann_benchmarks/runner.py", line 253, in run_docker
    _handle_container_return_value(return_value, container, logger)
  File "/tmp/ann-benchmarks/ann_benchmarks/runner.py", line 264, in _handle_container_return_value
    error_msg = return_value["Error"]
KeyError: 'Error'
```

It looks like `return_value` is `{'StatusCode': 0}` instead of `{'Error': None, 'StatusCode': 0}` (which I'm seeing with Docker 20 on Mac).


erikbern:
Thanks for fixing! I also saw this while I was running the benchmarks.

ankane:
No problem, and thanks!

PR #372, opened 2023-04-11
Fix the angular distance problem

SDU-L:


erikbern:
Thanks. You have to update the tests too. But I can do that.

PR #375, opened 2023-04-11
Added FastPQ

thomasahle:
Adds https://github.com/thomasahle/fast_pq to AnnBenchmarks

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -34,6 +34,7 @@ jobs:
           - elasticsearch
           - elastiknn
           - faiss
+          - fast_pq
           - flann
           - hnswlib
           - luceneknn
--- a/algos.yaml
+++ b/algos.yaml
@@ -213,7 +213,7 @@ float:
       constructor: FastPQ
       base-args: ["@metric"]
       run-groups:
-        annoy:
+        fast_pq:
           args: [[1, 2, 3]]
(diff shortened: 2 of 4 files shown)


erikbern:
let me know when this is ready to be merged!

thomasahle:
@erikbern What is actually right command to test run?
```
py run.py --docker-tag ann-benchmarks-fast_pq --max-n-algorithms 3 --runs 2 --dataset random-xs-20-angular --run-disabled --timeout 300
```
Seems to complain I don't have docker running.
Do I need to start my own docker daemon before I can run this?

thomasahle:
So adding
```
import sys, os
sys.path.append(os.getcwd() + '/fast_pq')
```
to `ann_benchmarks/algorithms/fast_pq.py` fixed my import errors, but I assume that's not the recommended approach.

thomasahle:
It's that the directory format is
```
fast_pq/
- fast_pq/
  - __init__.py
  - ...
- tests/
- setup.py
- ...
```
The PYTHONPATH should be such that `import fast_pq` imports the module (inner `fast_pq/`), not the outer directory.

erikbern:
Sure, but you don't need that hack if you just install it globally – which would be nice. That's how every other library works right now, and I think it's somewhat cleaner.

thomasahle:
That makes sense. I've updated the build system to work with `pip install .` so everything should be working now.

erikbern:
Awesome. I'll merge this!

erikbern:
Thanks for the contribution!

thomasahle:
Now just fingers crossed that it doesn't suck 🤞

PR #376, opened 2023-04-11
Fix Faiss CI failures

ankane:
CI currently fails when`faiss-gpu` is randomly selected to run (which is possible since CI uses `--run-disabled`).

```text
Trying to instantiate ann_benchmarks.algorithms.faiss_gpu.FaissGPU([4096, 1])
  File "run_algorithm.py", line 3, in <module>
    run_from_cmdline()
  File "/home/app/ann_benchmarks/runner.py", line 198, in run_from_cmdline
    run(definition, args.dataset, args.count, args.runs, args.batch)
  File "/home/app/ann_benchmarks/runner.py", line 101, in run
    algo = instantiate_algorithm(definition)
  File "/home/app/ann_benchmarks/algorithms/definitions.py", line 17, in instantiate_algorithm
    return constructor(*definition.arguments)
  File "/home/app/ann_benchmarks/algorithms/faiss_gpu.py", line 19, in __init__
    self._res = faiss.StandardGpuResources()
AttributeError: module 'faiss' has no attribute 'StandardGpuResources'
```

This fixes it, and incorporates two other changes (happy to split into separate PRs if desired):

1. Adds `-q` to reduce excessive `wget` output for Faiss build
2. Updates the job name on CI so it's easier to see what's running

Edit: Example failure - https://github.com/erikbern/ann-benchmarks/actions/runs/4662682480/jobs/8253299697

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -65,6 +65,9 @@ jobs:
     env:
       LIBRARY: ${{ matrix.library }}
       DATASET: ${{ matrix.dataset }}
+      RUN_OPTIONS: ${{ matrix.library != 'faiss' && '--run-disabled' || '' }}
+
+    name: ${{ matrix.library }} (${{ matrix.dataset }})
 
     steps:
     - uses: actions/checkout@v3 # Pull the repository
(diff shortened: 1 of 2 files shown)


ankane:
fwiw, I think all the remaining failures are due to libraries expecting a specific instruction set / AVX-512.

`milvus`

```
Illegal instruction (core dumped)
The command '/bin/sh -c python3 -c 'import pyknowhere'' returned a non-zero code: 132
```

`qsgngt`

```text
/home/app/qsg-union-3.0/lib/NGT/Index.h:Index:388: NGT::Index: Fatal Error!. Despite that This NGT library is built with AVX512, this CPU doesn't support AVX512. This CPU supoorts avx avx2 
```

`diskann`

```text
/usr/lib/gcc/x86_64-linux-gnu/11/include/avx512fintrin.h:305:1: error: inlining failed in call to 'always_inline' '__m512 _mm512_setzero_ps()': target specific option mismatch
```

PR #379, opened 2023-04-14
Redisearch

erikbern:
This is basically https://github.com/RedisAI/ann-benchmarks/tree/multiclient_tool rebased on current main


erikbern:
I think this got a bit too messy. I'm just going to copy the dockerfile and the algo script into this repo instead!

PR #380, opened 2023-04-14
update NGT parameters

masajiro:
The NGT parameters have been changed to fit the 2 h timeout. This is ready to merge


maumueller:
Thanks @masajiro!

PR #384, opened 2023-04-15
Added more config points to fast_pq

thomasahle:
Just to get some more recall in the 0.9+ range.


thomasahle:
Hm, this should be synced with the main repo. Not sure why the previous commits show up in the pull request. There doesn't seem to be any conflicts, and the "Changed Files" only contain the actual changes.

algos.yaml
```diff
@@ -215,7 +215,9 @@ float:
       run-groups:
         fast_pq:
           args: [[1, 2, 3]]
-          query-args: [[1, 2, 3, 4, 6, 8, 10, 13, 16, 20, 24, 28, 33, 38, 44, 50, 57, 64, 72, 80]]
+          # OEIS: A002984
```
erikbern, line 218: totally random side topic but maybe E series could be a good choice? https://en.wikipedia.org/wiki/E_series_of_preferred_numbers

erikbern:
let me know if you can't figure out the git issue and i can try to fix it

PR #389, opened 2023-04-17
update ngt parameters

masajiro:
The NGT parameters were updated again because they did not fit well within the 2 h timeout. Thank you for your trouble in advance.


erikbern:
Thanks!

PR #390, opened 2023-04-18
More query params for fast_pq

thomasahle:
Fixed git issues.
New values don't grow exponentially, but fast than the first "tier".


thomasahle:
Looks like the error is related to `glass`, not `fast_pq`.

PR #391, opened 2023-04-20
Added a skew factor to the number of clusters

thomasahle:
I talked to Matthijs Douze, and he recommended using a number of clusters that's slightly larger than sqrt(N).
This commit just adds three build arguments to do that.

### DIFF
--- a/algos.yaml
+++ b/algos.yaml
@@ -213,9 +213,26 @@ float:
       constructor: FastPQ
       base-args: ["@metric"]
       run-groups:
-        fast_pq:
-          # build_probes and skew_factor
-          args: [ [1, 2, 3], [1, 3, 9] ]
+        # Arguments are "build_probes" and "skew_factor"
+        # Time to build the index is mostly a factor of the skew_factor
+        # 1 | 578.4s
+        # 2 | 1033.9s
+        # 3 | 1410.6s
+        # 4 | 2420.7s (?)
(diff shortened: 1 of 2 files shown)


thomasahle:
@erikbern I want to make sure building the different parametrized indices don't take too long. 
Basically my fit/build consist of two tasks: (1) Running k-means on the dataset. (2) Adding each point to the `m` nearest clusters for some `m`.

The first task takes a long time 10-60 min depending on the number of cluster points.
The second task is quite fast. 1-3min.

The issue is, that with the current args vs query-args system, there's no way for me to cache the k-means clustering.
This means that when trying different values of `m`, the entire k-means process gets run from scratch.
Is there a way I can set up the configuration to prevent this problem? That would allow me to try more different `m` values.

thomasahle:
The ideal setting seems to be skew in [1, 4] and build_probes in [1, 9].
I could rewrite this pull-request using arg-groups, to just try one value for build-probes per skew. But if there's a way to cache the k-means, so I only have to do it 4 times, then I could try the whole product [1, 4] x [1, 9].

Figure_1

erikbern:
Seems a bit complex to refactor the benchmarks to support what you want to do – hopefully you can fit it into the current system!

thomasahle:
Sure, I've updated it with some more fine grained arg-groups. Hopefully they don't time out.

Issue #363, opened 2023-04-08
New run, April 2023


erikbern:
I just ran sift-128-euclidean and here are the results:

(edit: this chart is superseded by a later chart below)

I also ran glove-100-angular, but had a minor bug – see #362 – and I'm rerunning everything just to be on the safe side. Will post it later today.

Ran it on a [r6i.16xlarge](https://instances.vantage.sh/aws/ec2/r6i.16xlarge) with 31 runs in parallel – was able to run everything in just about 8 hours.

Note that these aren't final – I want to give library maintainers a chance to fix any issues. 

This plot is extremely crowded btw – any thoughts on algos we can remove? Or alternatively we could break it up into maybe "libraries" vs "servers" or something

erikbern:
glove-100-angular, 80% complete:

![image](https://user-images.githubusercontent.com/1027979/230752477-4cb2e394-56b7-486e-9124-f5aabd769879.png)

I'm killing this for now since it's $100/day+ to run. But will rerun soon!

erikbern:
Had an 1TB EBS volume but in retrospect that was kind of ridiculous – only needed 34GB up to this point!

WPJiang:
> glove-100-angular, 80% complete:
> 
> ![image](https://user-images.githubusercontent.com/1027979/230752477-4cb2e394-56b7-486e-9124-f5aabd769879.png)
> 
> I'm killing this for now since it's $100/day+ to run. But will rerun soon!
Hi  @erikbern ，in this test result, the recall of glove-100-angular  is close to zero, as is mentioned in  #367 . We checked the code and found that the calculation method of angular distance has changed. The normalized Euclidean distance currently used is order-preserving compared to angular distance,but their values are not equal. In current code，the distance threshold used to determine top-K is directly read from groundtruth (which is angular distance), while the distances of the top-K vectors are  normalized Euclidean distance.  We think this is cause of the nearly zero-recall problem. To solve the problem, we can take out the top-K ID of groundtruth and use the original vector to calculate the threshold for determination. Or use the original annbenchmarks angular distance calculation method.

erikbern:
Ohhh nice catch. I didn't realize changing the distance would matter.

I think the angular distance could be computed from dot products, is that right? Since

`dot(u, v) = |u| * |v| * cos(theta)`

then what we want is I think `theta = arccos( dot(u, v) / (|u| |v|) )`

is that right?

benbenwang1999:
[...]

PR #399, opened 2023-04-27
New results

erikbern:
See new plots


erikbern:
Will add a couple more datasets to this shortly!

thomasahle:
Oh no, looks like fast_pq/tinyknn regressed substantially since https://github.com/erikbern/ann-benchmarks/issues/363#issuecomment-1510390881 and is completely broken on mnist.
Are these new runs final? Or do we still have time to fix bugs?

thomasahle:
Is the run actually using the latest ann-benchmarks code? It seems the string fast_pq should have been replaced in the repo

erikbern:
> Is the run actually using the latest ann-benchmarks code? It seems the string fast_pq should have been replaced in the repo

I just realize I haven't re-run install.py in the last few days. Let me wipe the fast_pq data and re-run all benchmarks (will also wipe opensearchknn)

thomasahle:
> > Oh no, looks like fast_pq/tinyknn regressed substantially since [#363 (comment)](https://github.com/erikbern/ann-benchmarks/issues/363#issuecomment-1510390881) and is completely broken on mnist.
> > Are these new runs final? Or do we still have time to fix bugs?
> 
> If you fix it in the next few hours, then I can re-run it. But I want to tear down the r6.16xlarge instance soon (it's ~$100/day)

I found the problem. Just testing my fix now. Will push it asap.

thomasahle:
It's pushed.

WPJiang:
> Pushed latest results. Going to run gist-960-euclidean as well just to have a high dimensional dataset that's a bit larger. It will probably take 10h.

hi，@erikbern there is only one line -tinyknn in the new result of glove25, would you please check it?

erikbern:
> hi，@erikbern there is only one line -tinyknn in the new result of glove25, would you please check it?

You're right, I accidentally ran glove-25 only for tinyknn. Let me run it for all algos.

thomasahle:
It's strange. This is what I get when I spin up an r6i.8xlarge myself and run tinyknn on sift. Maybe I pushed the update too late, after you had already rerun install.py? Or maybe there's some other issue I can't think of...
sift-128-euclidean

erikbern:
It's possible the image didn't rebuild because of Docker layer caching. I didn't check closely. I'll rerun soon though.

thomasahle:
If the docker image times out during querying, is the data gathered still saved? Or is saving the data part of the job being run inside the docker container, so it gets lost? In the later case I can try to reduce the number of query_args to prevent timeouts.
It seems to me that the build-args/groups each have their own 7200s timeout, so I reducing them probably won't make a difference?

erikbern:
the process running inside the container saves the data to data that's mounted into the container from the outside – so if the container is killed after 2h, any data saved up to that point is kept

generally the 2h timeout happens during index building though, not afterwards during query processing. that step is usually pretty fast.

WPJiang:
> It's possible the image didn't rebuild because of Docker layer caching. I didn't check closely. I'll rerun soon though.
I think  this 
        for query_arguments in query_argument_groups:
            fn = get_result_filename(args.dataset, args.count, definition, query_arguments, args.batch)
            if args.force or not os.path.exists(fn):
                not_yet_run.append(query_arguments)

thomasahle:
> generally the 2h timeout happens during index building though

It seems some of the last args were taking more than half an hour, so I trimmed the list down. Now none of the rounds should take more than a minute or two.
https://github.com/erikbern/ann-benchmarks/pull/406

erikbern:
Merging this for now, but I'm planning to run this again in just a week or two. Will also polish the graphs a bit. But I don't want perfection to get in the way of getting something updated out. Won't promote this for now.

PR #403, opened 2023-04-28
Better qdrant params

generall:
Hi!

Sorry for late updates, just noticed that qdrant is included into the latest run.
In general I would say that comparing client-server vector DBs with pure embedded implementations of algorithms  might be misleading, but since other DBs are participating, we kinda have to as well.

In this PR I tried to do changes with the following goals:

- Minimize the overhead of data transfer: by default pydantic makes a lot of verification, that consumes CPU time. The PR changes it into raw gRPC data structures which doesn't have any validation
- Set the parameters for optimal RPS
- Try quantization

Hope you could re-run the latest iteration with those changes included.

Expected RPS increase with those changes: 2-3x

### DIFF
--- a/algos.yaml
+++ b/algos.yaml
@@ -552,8 +552,15 @@ float:
       base-args: ["@metric"]
       run-groups:
         default:
-          args: [[True, False]] #quantization
-          query-args: [[null, 8, 16, 32, 64, 128, 256]] #hnsw_ef
+          args: [
+            [True, False], #quantization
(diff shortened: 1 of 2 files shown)


erikbern:
Awesome. Thanks a lot for the contribution!

I think this is a bit too late for me to include in the benchmarks I'm publishing in #399, but I think I'm going to have to re-run everything in just a few weeks anyway. I don't intend to promote the results much until then.

generall:
Hi @erikbern, is there any ETA for the next run?

PR #406, opened 2023-04-28
Removed slow params from tinyknn

thomasahle:


erikbern:
would you mind trying to update this with latest main?

PR #407, opened 2023-04-29
Improve pgvector performance by ~2x

ankane:
Hi @erikbern, thanks for the updated benchmarks. Made two more improvements to pgvector:

1. Disabled compression and out-of-line storage
2. Set `-mprefer-vector-width=512`

Also, removed `lists = 10000` since it requires `maintenance_work_mem` to be higher.

### DIFF
--- a/algos.yaml
+++ b/algos.yaml
@@ -546,7 +546,7 @@ float:
       base-args: ["@metric"]
       run-groups:
         pgvector:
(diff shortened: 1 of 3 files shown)


erikbern:
Nice! Will re-run in maybe a week or two

ankane:
Great, thanks!

PR #410, opened 2023-05-04
update NGT parameters for GIST

masajiro:
A timeout didn't occur on r6i.xlarge for the previous parameters of NGT-QG. I think that it might be also possible for stuck containers or others to consume memory and eventually run out of it.


masajiro:
@erikbern
I would like you to merge this before re-running mentioned in your [comment](https://github.com/erikbern/ann-benchmarks/pull/403#issuecomment-1527802771).

PR #417, opened 2023-05-19
Adding read results script

ankitvij-db:
Script to read indexing stats that are stored as a json file with the name `{dataset}-{algo}-build-stats.json`

usage
`python3 read_results.py --dataset "glove-25-angular" --algo "opensearchknn" --path "<path to results>"`


ankitvij-db:
Accidental PR.

PR #434, opened 2023-07-02
feat: Add new dataset with OpenAI  embeddings for 1M DBpedia entities

KShivendu:
Add a new dataset with OpenAI embeddings for DBpedia entities.

This PR also introduces HuggingFace [datasets library](https://github.com/huggingface/datasets) which can help with loading any dataset on HuggingFace. This makes it easier for anyone to fork and benchmark against almost any public dataset :)

### DIFF
--- a/ann_benchmarks/datasets.py
+++ b/ann_benchmarks/datasets.py
@@ -475,16 +475,18 @@ def movielens10m(out_fn):
 def movielens20m(out_fn):
     movielens("ml-20m.zip", "ml-20m/ratings.csv", out_fn, ",", True)
 
-def dbpedia_entities_openai_1M(out_fn):
+def dbpedia_entities_openai_1M(out_fn, n = None):
     from sklearn.model_selection import train_test_split
     from datasets import load_dataset
     import numpy as np
 
     data = load_dataset("KShivendu/dbpedia-entities-openai-1M", split="train")
+    if n is not None and n >= 100_000:
+        data = data.select(range(n))
 
     embeddings = data.to_pandas()['openai'].to_numpy()
     embeddings = np.vstack(embeddings).reshape((-1, 1536))
-    
+
     X_train, X_test = train_test_split(embeddings, test_size=10_000, random_state=42)
 
     write_output(X_train, X_test, out_fn, "angular")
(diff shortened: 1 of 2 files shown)


ann_benchmarks/datasets.py
```diff
@@ -475,6 +475,22 @@ def movielens10m(out_fn):
 def movielens20m(out_fn):
     movielens("ml-20m.zip", "ml-20m/ratings.csv", out_fn, ",", True)
 
+def dbpedia_entities_openai_1M(out_fn):
+    from datasets import load_dataset
+    import numpy as np
+
+    data = load_dataset("KShivendu/dbpedia-entities-openai-1M", split="train")
```
erikbern, line 482: Where is this function defined?

ann_benchmarks/datasets.py
```diff
@@ -475,6 +475,22 @@ def movielens10m(out_fn):
...
+
+    embeddings = data.to_pandas()['openai'].to_numpy()
+    embeddings = np.vstack(embeddings).reshape((-1, 1536))
+
+    TEST_SIZE = 10_000
+
+    X_train = embeddings[TEST_SIZE:]
+    X_test = embeddings[:TEST_SIZE] # Take the first 10k as test set
```
erikbern, line 490: minor preference for using `sklearn.test_train_split` but not critical

ann_benchmarks/datasets.py
erikbern: oh nvm `datasets` is an external package. i was confused because the module has the same name :)

erikbern:
Nice! Are there bigger ones btw? We have a few datasets already that are around 1M vectors so it might be interesting to try something larger (like 3-10M)

KShivendu:
> Are there bigger ones btw?

While working on this [benchmark](https://twitter.com/NirantK/status/1674110063286571008?s=20) we didn't find any dataset with >=1536 dimensions. That's why I created one. We are planning to take this up to a 10M or even 100M scale in the upcoming weeks/months. I'll create PRs here when we do so :)

In the meantime, please note that this 1M `dbpedia-entities` dataset will take a lot of computing/RAM/time to run because of 1536 dimensions. One needs ~17GB RAM to run this with Qdrant and ~13GB for PGVector.

ann_benchmarks/datasets.py
KShivendu: Done

KShivendu: Done

erikbern:
Nice, thanks! I'll run this and will add the dataset to S3

PR #443, opened 2023-07-17
Feature/update luceneknn

benwtrent:
New release of Lucene takes advantage of some SIMD optimizations when doing vector search.

This PR updates to the newest Lucene version and updates the docker file so that the Panama Vector API can be utilized.

The following was done on: c3-standard-8 (Intel Sapphire Rapids)

mnist-784:
Log Scale improvement on 
image

Linear scale (same run): 
image


maumueller:
@benwtrent Sorry for missing this, I'm merging it now. Thanks for the contribution.

PR #445, opened 2023-07-23
add nndescent algorithm

brj0:
This adds support for [nndescent](https://github.com/brj0/nndescent), which is a C++ implementation of `pynndescent`.

Although single queries are slower compared to the original Python implementation, the build time (which includes the creation of a k-NN graph for the training data) on my machine shows a significant improvement. Here are the results of `ann-benchmarks` for **fashion-mnist-784-euclidean_10_euclidean**:

### Build time (s) tradeoff
Recall  | nndescent | pynndescent
--------|-----------|------------
0.9667  | 13.609    |
0.9693  | 14.184    |
0.9876  | 21.162    |
0.9877  |           |  81.0728
0.9997  | 42.560    | 111.5696
1.0000  | 51.341    | 125.3250

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -45,6 +45,7 @@ jobs:
           - luceneknn
           - milvus
           - mrpt
+          - nndescent
           - n2
           - nmslib
           - onng_ngt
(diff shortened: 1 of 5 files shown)


erikbern:
Nice!

PR #431, opened 2023-06-27
Documentation, Typing and Refactors

Jeadie:
## Changeset
 - Add typing to functions
 - Add docstrings to most functions
 - Change `Definition` from a `collections.namedtuple` to a `dataclass`
 - Simplify `ann_benchmarks.definitions.create_definitions_from_algorithm` by decomposing into separate functions.
 - Create a named tuple for Metric
    ```python
    class Metric(NamedTuple):
        distance: Callable[[np.ndarray, np.ndarray], float]
        distance_valid: Callable[[float], bool]
    ```
 - Simplify main function  in `ann_benchmarks/main.py` by moving filter functions (e.g. `filter_disabled_algorithms`) to separate python functions.
- NOTE: The above refactors should not alter the underlying logic of ann-benchmarks. 

## Motivation
I started working on specific changes to ann-benchmarks (e.g. adding new ANN algorithms, extending ann-benchmarks to be used as a performance testing library), and found the codebase hard to understand as a new user. This PR attempts to make the codebase easier to understand and well documented.

PR #447, opened 2023-07-26
Fix result caching by providing actual dataset name.

maumueller:
Fixes a regression introduced in #431 that led all runs re-run although results are present.


erikbern:
👍

PR #456, opened 2023-08-03
Get correct index size for pgvector

jkatz:
The size of the pgvector index comes from data stored in PostgreSQL; its not available in the memory context that the index is built in. This defines an ad hoc "get_memory_usage" function that lets the pgvector test load the size of the created index from the database.


maumueller:
Thanks @jkatz

PR #462, opened 2023-08-22
Fix dbpedia-openai-*k-angular datasets

ankane:
The current code generates datasets that are all 1M rows instead of the intended number.

https://stackoverflow.com/questions/6076270/lambda-function-in-list-comprehensions

### DIFF
--- a/ann_benchmarks/datasets.py
+++ b/ann_benchmarks/datasets.py
@@ -601,6 +601,6 @@ DATASETS: Dict[str, Callable[[str], None]] = {
 }
 
 DATASETS.update({
-    f"dbpedia-openai-{n//1000}k-angular": lambda out_fn: dbpedia_entities_openai_1M(out_fn, n)
+    f"dbpedia-openai-{n//1000}k-angular": lambda out_fn, i=n: dbpedia_entities_openai_1M(out_fn, i)
     for n in range(100_000, 1_100_000, 100_000)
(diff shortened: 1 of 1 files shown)


KShivendu:
Nice catch. I just validated this.

Before:
```python
>>> def dbpedia_entities_openai_1M(out_fn, n):
...     print(out_fn, n)
...
>>> d = {
    f"dbpedia-openai-{n//1000}k-angular": lambda out_fn: dbpedia_entities_openai_1M(out_fn, n)
    for n in range(100_000, 1_100_000, 100_000)
}
>>> d['dbpedia-openai-100k-angular']('abc')
abc 1000000
```

After:
```python
>>> def dbpedia_entities_openai_1M(out_fn, n):
...     print(out_fn, n)
...
>>> d = {
    f"dbpedia-openai-{n//1000}k-angular": lambda out_fn, i=n: dbpedia_entities_openai_1M(out_fn, i)
    for n in range(100_000, 1_100_000, 100_000)
}
>>> d['dbpedia-openai-100k-angular']('abc')
abc 100000
```

@erikbern please approve if it looks good to you.

KShivendu:
@maumueller requesting your review on this :)

maumueller:
nice catch indeed, thanks @ankane 

@erikbern could you publish these datasets, thanks?

PR #463, opened 2023-08-31
Use HNSW for pgvector

ankane:
Added in pgvector 0.5.0


maumueller:
Thanks!

ankane:
Thanks @maumueller!

PR #472, opened 2023-10-15
Disable a few algos

erikbern:
These are algos that are either chronically bad or dominated by some other flavor within the same family

Sad to see you go, Annoy, but the time has come

### DIFF
--- a/ann_benchmarks/algorithms/annoy/config.yml
+++ b/ann_benchmarks/algorithms/annoy/config.yml
@@ -2,7 +2,7 @@ bit:
   hamming:
   - base_args: ['@metric']
     constructor: Annoy
-    disabled: false
+    disabled: true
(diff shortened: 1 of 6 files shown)


masajiro:
@erikbern 

> These are algos that are either chronically bad or dominated by some other flavor within the same family

NGT includes QG, ONNG, and PANNG. QG is a quantization and graph-based method. ONNG and PANNG are graph-based methods. Moreover, ONNG and PANNG use different methods of graph construction. If one of these should be disabled, PANNG should be disabled instead of ONNG since PANNG is worse than the others.

KShivendu:
Surprised to see that Qdrant is getting disabled. Would like to understand why are we doing this?

KShivendu:
@erikbern why are we disabling Qdrant?

PR #488, opened 2024-01-31
Update pgvector loading method to use binary format

jkatz:
Particularly on large vector types, the pgvector module was spending significant time on converting floating point values to ASCII before being transmitted to the PostgreSQL server. This changes keeps the format in binary, reducing overhead. One test demonstrated a 63% reduction in load time, which would have an impact on the overall "build" time as reported by this benchmark.


maumueller:
Thanks!

PR #491, opened 2024-03-01
pgvector updates

ankane:
Hi, this PR makes two updates to pgvector:

1. Fixes `could not resize shared memory segment` error with pgvector 0.6.0 and Docker by explicitly disabling parallel index builds (`max_parallel_maintenance_workers` change)
2. Updates Postgres from 14 to 16 (which requires the `GRANT` change)


maumueller:
Thanks!

PR #496, opened 2024-03-16
fix docker tag for Faiss HNSW

koron:
Fixed an error that occurred when executing in an environment where the docker image for faiss_hnsw existed and the docker image for faiss did not exist.

This will be failed:

```console
# In the environment where ANN Benchmarks is executed for the first time

$ python install.py --algorithm faiss_hnsw

$ python run.py --algorithm 'hnsw(faiss)
```


maumueller:
You are right, that was indeed a bug. Thanks for the contribution.

(I wonder why we didn't consolidate all FAISS implementations into a single module.)

PR #505, opened 2024-03-29
fix error reporting of runner

tureba:
When an error is reported, container.logs() returns an iterator over lines of bytes. The decode() function should apply to each line, not to the iterator, otherwise we get no logs, but an error message saying there is no decode() function in the iterator.

This change follows the same pattern as stream_logs() in the same file.


erikbern:
thanks!

Issue #420, opened 2023-05-26
Hamming distance graphs seem to be empty


DmitryKey:
https://ann-benchmarks.com/sift-256-hamming_10_hamming.html
<img width="905" alt="Screenshot 2023-05-26 at 9 36 48" src="https://github.com/erikbern/ann-benchmarks/assets/123553/00e8cd8b-619c-4220-a68f-fcaf554b3bcb">

Same concerns Recall / Build time, Recall / Index size, ..

The only populated graph is "Relative Error/Queries per second (1/s)".

PR #508, opened 2024-04-06
Fix recall for Hamming distance

ankane:
Both `sift-256-hamming` and `word2bits-800-hamming` always report 0 recall, as the distances in the HDF5 files are floats between 0 and 1 rather than the Hamming distance. Multiplying by the dataset dimensions fixes it.

This issue is likely the cause of #420.

### DIFF
--- a/ann_benchmarks/distance.py
+++ b/ann_benchmarks/distance.py
@@ -27,7 +27,7 @@ class Metric(NamedTuple):
 
 metrics = {
     "hamming": Metric(
-        distance=lambda a, b: np.sum(a.astype(np.bool_) ^ b.astype(np.bool_)),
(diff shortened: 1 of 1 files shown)


ankane:
Another approach would be to change Hamming distance to use `mean`.

```diff
 metrics = {
     "hamming": Metric(
-        distance=lambda a, b: np.sum(a.astype(np.bool_) ^ b.astype(np.bool_)),
+        distance=lambda a, b: np.mean(a.astype(np.bool_) ^ b.astype(np.bool_)),
         distance_valid=lambda a: True
     ),
```

maumueller:
Thanks!

ankane:
Thanks @maumueller!

PR #512, opened 2024-04-16
Add IVF bench for PgVector extension

C0rWin:
Currently, the pgvector benchmark looks only at the HNSW index, missing the second index type, IVFFlat. This commit also extends the pgvector benchmark and the ability to execute the IVFFlat index. Though I understand HNSW might have superior performance, it's still interesting to have the ability to compare it with another index type.

### DIFF
--- a/ann_benchmarks/algorithms/pgvector/config.yml
+++ b/ann_benchmarks/algorithms/pgvector/config.yml
@@ -5,7 +5,7 @@ float:
     disabled: false
     docker_tag: ann-benchmarks-pgvector
     module: ann_benchmarks.algorithms.pgvector
-    name: pgvector
+    name: pgvector_hnsw
     run_groups:
       M-16:
         arg_groups: [{M: 16, efConstruction: 200}]
@@ -15,3 +15,12 @@ float:
         arg_groups: [{M: 24, efConstruction: 200}]
         args: {}
         query_args: [[10, 20, 40, 80, 120, 200, 400, 800]]
+  - base_args: ['@metric']
+    constructor: PGVectorIVF
+    disabled: false
+    docker_tag: ann-benchmarks-pgvector
+    module: ann_benchmarks.algorithms.pgvector
+    name: pgvector_ivf
+    run_groups:
+      base:
+        args: [[32, 64, 128, 256, 512, 1024]]
--- a/ann_benchmarks/algorithms/pgvector/module.py
+++ b/ann_benchmarks/algorithms/pgvector/module.py
@@ -62,3 +62,55 @@ class PGVector(BaseANN):
 
     def __str__(self):
         return f"PGVector(m={self._m}, ef_construction={self._ef_construction}, ef_search={self._ef_search})"
+
+
+class PGVectorIVF(BaseANN):
+    def __init__(self, metric, n_list):
+        self._metric = metric
+        self._n_list = n_list
+        self._cur = None
+
+        if metric == "angular":
+            self._query = "SELECT id FROM items ORDER BY embedding <=> %s LIMIT %s"
+        elif metric == "euclidean":
+            self._query = "SELECT id FROM items ORDER BY embedding <-> %s LIMIT %s"
+        else:
+            raise RuntimeError(f"unknown metric {metric}")
+
+    def fit(self, X):
+        subprocess.run("service postgresql start", shell=True, check=True, stdout=sys.stdout, stderr=sys.stderr)
+        conn = psycopg.connect(user="ann", password="ann", dbname="ann", autocommit=True)
+        pgvector.psycopg.register_vector(conn)
+        cur = conn.cursor()
+        cur.execute("DROP TABLE IF EXISTS items")
+        cur.execute("CREATE TABLE items (id int, embedding vector(%d))" % X.shape[1])
+        cur.execute("ALTER TABLE items ALTER COLUMN embedding SET STORAGE PLAIN")
+        print("copying data...")
+        with cur.copy("COPY items (id, embedding) FROM STDIN WITH (FORMAT BINARY)") as copy:
+            copy.set_types(["int4", "vector"])
+            for i, embedding in enumerate(X):
+                copy.write_row((i, embedding))
+        print("creating index...")
+        if self._metric == "angular":
+            cur.execute(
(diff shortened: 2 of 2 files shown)


ann_benchmarks/algorithms/pgvector/config.yml
```diff
@@ -5,7 +5,7 @@ float:
     disabled: false
     docker_tag: ann-benchmarks-pgvector
     module: ann_benchmarks.algorithms.pgvector
-    name: pgvector
+    name: pgvector_hnsw
```
C0rWin, line 8: renamed only to make it clear while comparing with ivfflat.

C0rWin:
```
_________________________________ test_hamming _________________________________

    def test_hamming():
        dist = metrics["hamming"].distance
    
        p = numpy.array([1, 1, 0, 0], dtype=numpy.bool_)
        q = numpy.array([1, 0, 0, 1], dtype=numpy.bool_)
>       assert dist(p, q) == pytest.approx(2)
E       assert 0.5 == 2 ± 2.0e-06
E         comparison failed
E         Obtained: 0.5
E         Expected: 2 ± 2.0e-06

test/distance_test.py:[21](https://github.com/erikbern/ann-benchmarks/actions/runs/8704604263/job/23873209353?pr=512#step:5:22): AssertionError
=========================== short test summary info ============================
```

not sure whenever and how is PR related to the UT failure.

C0rWin:
> ```
> _________________________________ test_hamming _________________________________
> 
>     def test_hamming():
>         dist = metrics["hamming"].distance
>     
>         p = numpy.array([1, 1, 0, 0], dtype=numpy.bool_)
>         q = numpy.array([1, 0, 0, 1], dtype=numpy.bool_)
> >       assert dist(p, q) == pytest.approx(2)
> E       assert 0.5 == 2 ± 2.0e-06
> E         comparison failed
> E         Obtained: 0.5
> E         Expected: 2 ± 2.0e-06
> 
> test/distance_test.py:[21](https://github.com/erikbern/ann-benchmarks/actions/runs/8704604263/job/23873209353?pr=512#step:5:22): AssertionError
> =========================== short test summary info ============================
> ```
> 
> not sure whenever and how is PR related to the UT failure.

tested with the previous commit, getting the same error.... I assume there are some issues with the main branch.

C0rWin:
https://github.com/erikbern/ann-benchmarks/pull/513 should expected to fix the UT issue.

jkatz:
cc @ankane 

The IVFFlat testing for pgvector was explicitly removed in https://github.com/erikbern/ann-benchmarks/pull/463 -- Andrew & I had discussed this quite a bit and opted for the simpler approach of only having HNSW present in the ANN Benchmarks suite. Given this is the primary indexing method + de facto choice for pgvector, it makes sense to just have the single method present.

If we did decide to do too, I would opt to break it out as "pgvector_ivfflat" and keep the HNSW implementation as "pgvector"

ann_benchmarks/algorithms/pgvector/module.py
```diff
@@ -62,3 +62,55 @@ def get_memory_usage(self):
 
     def __str__(self):
         return f"PGVector(m={self._m}, ef_construction={self._ef_construction}, ef_search={self._ef_search})"
+
+
+class PGVectorIVF(BaseANN):
```
jkatz, line 67: If this is to be added, I'd recommend using the original implementation, though adding the binary load found in https://github.com/erikbern/ann-benchmarks/pull/488:

https://github.com/erikbern/ann-benchmarks/tree/a884247a83db918d4f19c82048293c8fce65b30b/ann_benchmarks/algorithms/pgvector

C0rWin:
> cc @ankane
> 
> The IVFFlat testing for pgvector was explicitly removed in #463 -- Andrew & I had discussed this quite a bit and opted for the simpler approach of only having HNSW present in the ANN Benchmarks suite. Given this is the primary indexing method + de facto choice for pgvector, it makes sense to just have the single method present.
> 
> If we did decide to do too, I would opt to break it out as "pgvector_ivfflat" and keep the HNSW implementation as "pgvector"

Here is the thing: I do not think this is a de facto method since there might be workloads where you'd like to compromise the accuracy in favor of speed; having this benchmark might help to shed some light on differences between these two. Obviously, this is only IMHO.

ann_benchmarks/algorithms/pgvector/module.py
C0rWin: I wasn't aware it was part of the ann_benchmark workload. It makes sense to bring back the original implementation, but I don't know why it was removed in the first place; I am trying to learn the tradeoffs presented by both algorithm implementations and find it extremely useful.

jkatz:
> > cc @ankane
> > The IVFFlat testing for pgvector was explicitly removed in #463 -- Andrew & I had discussed this quite a bit and opted for the simpler approach of only having HNSW present in the ANN Benchmarks suite. Given this is the primary indexing method + de facto choice for pgvector, it makes sense to just have the single method present.
> > If we did decide to do too, I would opt to break it out as "pgvector_ivfflat" and keep the HNSW implementation as "pgvector"
> 
> Here is the thing: I do not think this is a de facto method since there might be workloads where you'd like to compromise the accuracy in favor of speed; having this benchmark might help to shed some light on differences between these two. Obviously, this is only IMHO.

I'm in a position where I can make some more data-driven opinions on this, and from the data and conversations with pgvector users, the typical approach now is to go "HNSW-first." There are cases where I talk to users who are using IVFFlat and are very happy with their results, their workloads tend to be more static (all vectors are available at indexing time), whereas most use I'm seeing these days tends to be iterative (vectors are continuosuly added to the data set).

While we were evaluating pgvector's HNSW implementation, I did [publish benchmarks (using an adapted ANN Benchmark) on the two methods](https://jkatz05.com/post/postgres/pgvector-hnsw-performance/). I just happened to be doing some comparative analysis on the upcoming pgvector release (which I'll publish), but the performance/recall ratio gap between pgvector's HNSW + IVFFlat implementation has grown, and IVFFlat's biggest advantage (index build time) has dwindled or been superceded. In the coming days I'll have additional data on this point.

I can understand from an evaluation standpoint where it may be helpful to compare the two methods, but from a simplicity and usage standpoint (and where the referenced changes to the module game), it makes more sense to compare the HNSW implementation. Maybe the compromise is including the code to execute an evaluation with IVFFlat, but have it `disabled` by default?

PR #513, opened 2024-04-16
Fix hamming distance test

C0rWin:


ankane:
Hi @C0rWin, check out #508.

C0rWin:
> Hi @C0rWin, check out #508.

I might be missing something, but master still has this issue:

https://github.com/erikbern/ann-benchmarks/blob/75043ab482d91afae82ac4033cbdd98997121d58/ann_benchmarks/distance.py#L30

ankane:
That code is correct for the benchmarks to show the correct recall with the hosted datasets (see the linked issue or test it yourself).

PR #522, opened 2024-05-17
 add `--count` and `--batch` args for data_export.py

jiangyinzuo:
also convert `X` to numpy.array in glove dataset function

### DIFF
--- a/ann_benchmarks/results.py
+++ b/ann_benchmarks/results.py
@@ -84,7 +84,7 @@ def load_all_results(dataset: Optional[str] = None,
     Yields:
         tuple: A tuple containing properties as a dictionary and an h5py file object.
     """
-    for root, _, files in os.walk(build_result_filepath(dataset, count)):
+    for root, _, files in os.walk(build_result_filepath(dataset, count, batch_mode=batch_mode)):
         for filename in files:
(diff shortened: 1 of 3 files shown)


maumueller:
Thanks @jiangyinzuo. What is the actual use case for adding these parameters? The idea of `data_export.py` is to export all the results that are found. Everything else was supposed to be done in post-processing the csv files.

ann_benchmarks/results.py
```diff
@@ -84,7 +84,7 @@ def load_all_results(dataset: Optional[str] = None,
     Yields:
         tuple: A tuple containing properties as a dictionary and an h5py file object.
     """
-    for root, _, files in os.walk(build_result_filepath(dataset, count)):
+    for root, _, files in os.walk(build_result_filepath(dataset, count, batch_mode=batch_mode)):
```
jiangyinzuo, line 87: If `data_export.py` is expected to export all the results that are found, then here `build_result_filepath` should return all the sub-directories.

However, in `def build_result_filepath(...):`

https://github.com/erikbern/ann-benchmarks/blob/3982de0c267c9e3de65044b5c55f143917df7023/ann_benchmarks/results.py#L11-L38

If `count` and `batch_mode` are `None`, directories of specific counts and batch mode will be ignored.

PR #527, opened 2024-05-23
Use custom definitions for algorithms

tymmale:
Hi! Great benchmark that you have here.

I noticed that the value of the `--definitions` argument is only used when listing all available algorithms. Correct me if I'm wrong, but I wanted to pass my own algorithm configurations, which are stored elsewhere, by using `--definitions`. This doesn't work because the program uses the default path `ann-benchmarks/algorithms`.


maumueller:
Thanks @tymmale, this is indeed what `--definitions` was supposed to do. I'm waiting for the CI and will then merge this.

PR #538, opened 2024-08-01
Add ParlayANN to ann-benchmarks

magdalendobson:
ParlayANN implements a variety of algorithms as described in the paper: https://dl.acm.org/doi/10.1145/3627535.3638475  This pull request uses the ParlayANN version of Vamana (with some significant changes from the original).  It was designed for large data sets and high parallelism, but works well on smaller graphs and a single thread.

It does not use any product quantization, but does use scalar quantization, and for large dimensions (just GIST is large enough for this suite) uses  the Johnson-Lindenstrauss transform to reduce dimensionality.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -54,6 +54,7 @@ jobs:
           - onng_ngt
           - opensearchknn
           - panng_ngt
+          - parlayann
           - pg_embedding
           - pgvector
           - pgvecto_rs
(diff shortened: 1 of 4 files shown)


erikbern:
Nice! reminds me I need to re-run the benchmarks

magdalendobson:
Looking forward to seeing the results! Thanks so much!

erikbern:
thanks for the contribution!

PR #542, opened 2024-08-22
Track move of vespa rpm repository from Fedora Copr to Cloudsmith.

toregge:


maumueller:
Thank you for the contribution, @toregge

PR #543, opened 2024-08-31
Update python docker dependency to 7.1.0 to fix 'Not supported URL scheme http+docker' exception

alexklibisz:
When running on Ubuntu 22.04 LTS with 3.10.8, I'm seeing an error like below. Upgrading to the latest version of the docker library makes the error go away.

```
Traceback (most recent call last):                                                                                                                                                                                                                                                                                                                                         
  File "/home/ubuntu/elastiknn/ann-benchmarks/ann-benchmarks/venv/lib/python3.10/site-packages/requests/adapters.py", line 633, in send                                                                                                                                                                                                                                    
    conn = self.get_connection_with_tls_context(                                                                                                                                                                                                                                                                                                                           
  File "/home/ubuntu/elastiknn/ann-benchmarks/ann-benchmarks/venv/lib/python3.10/site-packages/requests/adapters.py", line 489, in get_connection_with_tls_context                                                                                                                                                                                                         
    conn = self.poolmanager.connection_from_host(                                                                                                                                                                                                                                                                                                                          
  File "/home/ubuntu/elastiknn/ann-benchmarks/ann-benchmarks/venv/lib/python3.10/site-packages/urllib3/poolmanager.py", line 303, in connection_from_host                                                                                                                                                                                                                  
    return self.connection_from_context(request_context)                                                                                                                                                                                                                                                                                                                   
  File "/home/ubuntu/elastiknn/ann-benchmarks/ann-benchmarks/venv/lib/python3.10/site-packages/urllib3/poolmanager.py", line 325, in connection_from_context                                                                                                                                                                                                               
    raise URLSchemeUnknown(scheme)                                                                                                                                                                                                                                                                                                                                         
urllib3.exceptions.URLSchemeUnknown: Not supported URL scheme http+docker                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                           
During handling of the above exception, another exception occurred:                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                           
Traceback (most recent call last):                                                                                                                                                                                                                                                                                                                                         
  File "/home/ubuntu/elastiknn/ann-benchmarks/ann-benchmarks/venv/lib/python3.10/site-packages/docker/api/client.py", line 214, in _retrieve_server_version                                                                                                                                                                                                                
    return self.version(api_version=False)["ApiVersion"]                                                                                                                                                                                                                                                                                                                   
[...]

PR #544, opened 2024-08-31
Update Elastiknn to latest release and latest benchmark parameters

alexklibisz:
This also includes the change from #543, which was necessary to get this to run.


maumueller:
Thanks for the contribution, @alexklibisz

PR #548, opened 2024-10-18
Support for MariaDB database

HugoWenTD:
As MariaDB is building Vector support, introducing new module for benchmarking against MariaDB database server.

**Mark the current implementation as a draft as it is based on MariaDB preview release. We expect MariaDB to release Vector support in innovation release and then syntax and configuration settings may be updated.**

### DIFF
--- a/ann_benchmarks/algorithms/mariadb/module.py
+++ b/ann_benchmarks/algorithms/mariadb/module.py
@@ -1,0 +1,309 @@
+import getpass
+import glob
+import os
+import shutil
+import signal
+import struct
+import subprocess
+import sys
+import tempfile
+import time
+
+import mariadb
+
+from ..base.module import BaseANN
+
+def vector_to_hex(v):
+    binary_data = bytearray(v.size * 4)
+    for index, f in enumerate(v):
+        struct.pack_into('f', binary_data, index * 4, f)
+    return binary_data
+
+class MariaDB(BaseANN):
+
+    def __init__(self, metric, method_param):
+        self._test_time = time.strftime("%Y-%m-%d-%H-%M-%S", time.localtime())
+        self._metric = metric
+        self._m = method_param['M']
+        self._ef_construction = method_param['efConstruction']
+        self._cur = None
+        self._perf_proc = None
+        self._perf_records = []
+        self._perf_stats = []
+
+
+        if metric == "angular":
+            raise RuntimeError(f"Angular metric is not supported.")
+        elif metric == "euclidean":
+            # euclidean is the current default and only distance metric supported by MariaDB
+            pass
+        else:
+            raise RuntimeError(f"unknown metric {metric}")
+        
+        self.prepare_options()
+        self.initialize_db()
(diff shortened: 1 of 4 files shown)


maumueller:
Thanks @HugoWenTD. Please make sure to add a [test case to the CI as well](https://github.com/erikbern/ann-benchmarks/blob/main/.github/workflows/benchmarks.yml).

erikbern:
Are you planning to finish this @HugoWenTD ?

HugoWenTD:
> Are you planning to finish this @HugoWenTD ?

Yes Erik, I'll work on it this week.

erikbern:
Ok, nice!

ann_benchmarks/algorithms/mariadb/module.py
```diff
@@ -0,0 +1,124 @@
...
+            # Insert when batch is full
+            if len(batch) >= batch_size:
+                self._cur.executemany(self._sql_insert, batch)
+                batch.clear()
+        
+        # Insert remaining records in final batch
+        if batch:
+            self._cur.executemany(self._sql_insert, batch)
+    
+        insert_time = time.time() - start_time
```
cvicentiu, line 92: Perhaps it would make sense to time just the "executemany" calls, not the python overhead of creating the batch?

If one is to compare these numbers to other DBs, it should time the operation the same. Is it *with* python overhead, or *without* in other DB's cases?

erikbern: We don't really benchmark insertions & building the index, so I don't think it's material!

erikbern:
This is failing in CI for some reason, any idea why? https://github.com/erikbern/ann-benchmarks/actions/runs/14016901547/job/39890033550?pr=548

ann_benchmarks/algorithms/mariadb/module.py
HugoWenTD: Yeah, it's just a print, but I'll see if can change it.

HugoWenTD:
> This is failing in CI for some reason, any idea why? https://github.com/erikbern/ann-benchmarks/actions/runs/14016901547/job/39890033550?pr=548

Weird. Looks it's failing with `--batch` option. I'll look into it.
```
ERROR - double free or corruption (!prev)
```

HugoWenTD:
> This is failing in CI for some reason, any idea why? https://github.com/erikbern/ann-benchmarks/actions/runs/14016901547/job/39890033550?pr=548

I suspect it's related to the ThreadPool and mariadb cursor.

```
    def batch_query(self, X: numpy.array, n: int) -> None:
        pool = ThreadPool()
        self.res = pool.map(lambda q: self.query(q, n), X)
```

```
    def query(self, v, n):
        self._cur.execute(self._sql_search, (self.vector_to_hex(v), n))

        return [id for id, in self._cur.fetchall()]
```

HugoWenTD:
Updated the batch_query. It should work now.

PR #559, opened 2024-12-20
Add multimodal dataset based on COCO text-image pairs

fabiocarrara:
This PR adds two ANN datasets drived from the [COCO](https://cocodataset.org/) text-image pairs dataset:

1. **COCO Text-to-Image Multimodal Dataset (`coco-t2i-512-angular`)**:
   - Text is used as queries, and images comprise the search set.
   - This dataset presents a challenge due to the distribution data shift between queries and the search set, with the 100 nearest neighbors of queries having a cosine similarity of 0.30 +- 0.02.
2. **COCO Image-to-Image Intra-modal Dataset (`coco-i2i-512-angular`)**:
   - Images are used as both queries and search set.
   - This dataset does not exhibit the distribution shift and can serve as a reference, sharing the same datapoints as the `t2i` dataset.

**Extraction Process**
Features vectors are the CLS output token of the [OpenAI's CLIP](https://github.com/openai/CLIP) with ViT-B/16 architecture (512 dimensions) of the visual or textual encoder. Thanks to @mesnico for performing the extraction.

**Split definition**
Based on Karpathy's split of COCO 2014:

 - The search sets include 118,287 vectors extracted from training set images.
 - Queries:
   - Visual (`i2i`): 5,000 vectors from validation set images.
   - Textual (`t2i`): 5,000 vectors from the first caption (out of the five available) of validation set images.

@maumueller: Lucia (@vadicamo) told me you were searching for a multimodal dataset for the SISAP indexing challenge. If it's still needed, check whether those could be a good fit. Let me know if you'd like more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,8 @@ We have a number of precomputed data sets in HDF5 format. All data sets have bee
 | [NYTimes](https://archive.ics.uci.edu/ml/datasets/bag+of+words)   |        256 |    290,000 |    10,000 |       100 | Angular   | [HDF5](http://ann-benchmarks.com/nytimes-256-angular.hdf5) (301MB)         |
 | [SIFT](http://corpus-texmex.irisa.fr/)                           |        128 |  1,000,000 |    10,000 |       100 | Euclidean | [HDF5](http://ann-benchmarks.com/sift-128-euclidean.hdf5) (501MB)          |
 | [Last.fm](https://github.com/erikbern/ann-benchmarks/pull/91)     |         65 |    292,385 |    50,000 |       100 | Angular   | [HDF5](http://ann-benchmarks.com/lastfm-64-dot.hdf5) (135MB)               |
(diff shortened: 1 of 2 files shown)


fabiocarrara:
Some (partial) results on `coco-t2i-512-angular`:

coco-t2i-512-angular

maumueller:
Thanks for the contribution, @fabiocarrara! I particularly like that the order of algorithms is changing between the different datasets. 

What I'm worried about in the current setup is that we are trusting your hdf5 computation. Is it possible to change the pipeline in such a way that only the vectors are downloaded and ann-benchmarks does the train/test-split and computes the groundtruth? For example, even the openblas baseline doesn't seem to achieve recall 1 on the image-2-image version.

maumueller:
Thank you, @fabiocarrara. Before I merge this, could you please make sure that the hdf5 file that you are pointing to in the table is the one that is produced by running `create_dataset.py`? A potential inconsistency could lead to a lot of confusion :-)

@erikbern Can you upload the two hdf5 files to ann-benchmarks.com, please?

maumueller:
Thanks!

Issue #560, opened 2024-12-20
Unable to download mnist-784-euclidean


mdcallag:
I can download fashion-mnist-784-euclidean, sift-128-euclidean and gist-960-euclidean. 
But when I try to download mnist-784-euclidean I get an error.
```
python3 run.py --dataset mnist-784-euclidean

downloading https://ann-benchmarks.com/mnist-784-euclidean.hdf5 -> data/mnist-784-euclidean.hdf5...
Cannot download https://ann-benchmarks.com/mnist-784-euclidean.hdf5
Creating dataset locally
downloading http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz -> mnist-train.gz...
Traceback (most recent call last):
  File "/home/mdcallag/git/ann_vuvova/ann_benchmarks/datasets.py", line 55, in get_dataset
    download(dataset_url, hdf5_filename)
  File "/home/mdcallag/git/ann_vuvova/ann_benchmarks/datasets.py", line 21, in download
    urlretrieve(source_url, destination_path)
  File "/usr/lib/python3.10/urllib/request.py", line 241, in urlretrieve
    with contextlib.closing(urlopen(url, data)) as fp:
  File "/usr/lib/python3.10/urllib/request.py", line 216, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib/python3.10/urllib/request.py", line 525, in open
    response = meth(req, response)
  File "/usr/lib/python3.10/urllib/request.py", line 634, in http_response
    response = self.parent.error(
  File "/usr/lib/python3.10/urllib/request.py", line 563, in error
    return self._call_chain(*args)
  File "/usr/lib/python3.10/urllib/request.py", line 496, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.10/urllib/request.py", line 643, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 403: Forbidden
```

I also get an error if I try to fetch that file via wget
```
$ wget  http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz

--2024-12-20 23:52:05--  http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz
Resolving yann.lecun.com (yann.lecun.com)... 216.165.22.6
Connecting to yann.lecun.com (yann.lecun.com)|216.165.22.6|:80... connected.
[...]

PR #561, opened 2024-12-23
Fix downloads

erikbern:
This resolves https://github.com/erikbern/ann-benchmarks/issues/560


erikbern:
Clearly we need to clean up a bunch of failing ones

PR #562, opened 2024-12-25
Modify the string of hnswlib's parameters to add efQuery

guangzegu:
When exporting the CSV results of hnswlib, the parameters in the file do not include the ef value from the query arguments. This pull request modifies self.name to include it, facilitating performance comparison.


maumueller:
This is an excellent observation, @guangzegu. Thank you for your contribution.

Issue #563, opened 2024-12-26
Allow specifying database server connection credentials


michaelbautin:
For example, when running pgvector against a non-Docker PostgreSQL server, we need to be able to specify database host/port/username/database name/password.

PR #564, opened 2024-12-27
Add ability to specify Postgres connection credentials

michaelbautin:
- Allow specifying PostgreSQL credentials when running against an external instance of PostgreSQL not managed by the benchmark. 
- Allow avoiding an attempt to start PostgreSQL service with systemctl, because it might be managed in a different way. 
- Create the `vector` PostgreSQL extension automatically if it does not exist.

Fixes: #563

### DIFF
--- a/ann_benchmarks/algorithms/pgvector/module.py
+++ b/ann_benchmarks/algorithms/pgvector/module.py
@@ -57,7 +57,7 @@ class PGVector(BaseANN):
             else:
                 print("vector extension does not exist, creating")
                 cur.execute("CREATE EXTENSION vector")
-        
+
     def fit(self, X):
         psycopg_connect_kwargs: Dict[str, Any] = dict(
             autocommit=True,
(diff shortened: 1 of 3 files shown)


maumueller:
Thanks for your contribution, @michaelbautin. I'm a bit skeptical about the proposed solution, because it makes this very pronounced from an API point of view, while it only applies to a very small subset of the algorithms. 

Can you achieve the same functionality by specifying these credentials in the config file https://github.com/erikbern/ann-benchmarks/blob/main/ann_benchmarks/algorithms/pgvector/config.yml? If I remember correctly, you can provide it in the `base_args` entry which is propagated to the constructor of the class.

michaelbautin:
@maumueller Thank you for your feedback. I've reworked the PR to avoid introducing generic command-line parameters, and instead made it controllable by Postgres-specific environment variables. Also I've added automatic pgvector extension creation and a way to avoid attempting to start PostgreSQL service if it is managed in a different way.

Cc @erikbern

michaelbautin:
@maumueller yes, my use case is that the Postgres server is not managed by ann-benchmarks scripts. In my experiments I frequently need a Postgres server set up in a very specific way in a cloud environment.

maumueller:
Alright, I can merge this if you agree that it provides the necessary functionality. One last thing:  Do you think a short comment in the `module.py` would help others find out about this feature? (And if yes, could you please add one? Thanks!)

michaelbautin:
@maumueller : sounds great, thank you! I will add the comment shortly.

michaelbautin:
@maumueller I've added the requested comment. Thank you!

maumueller:
Thank you for the contribution, @michaelbautin!

PR #567, opened 2025-01-16
Update qsgngt

SDU-L:


maumueller:
Thanks for the contrubution, @SDU-L.

PR #578, opened 2025-03-17
Upgrade to Ubuntu 24.04 and Python 3.12

erikbern:
Prepping for a new run and this is various updates to get it working on Ubuntu 24.04 and Python 3.12


erikbern:
So bad news is that like 40% of algorithms are still broken, even after me trying to fix them.

Good news is maybe that's fine? Honestly a lot of these are super brittle code bases that haven't been updated, and I think endorsing them from this project is maybe bad. And it will make the plots a bit easier to read.

erikbern:
Open to staying on Python 3.10 if people feel strongly about it. There are some changes that I think is causing a few compilation failures. And a few of the algorithms have wheels prepared for 3.10 but not 3.12.

erikbern:
I think I'll close this and start over with 22.04 and 3.10

We should upgrade to more modern software but it's still not super behind, so maybe it's fine to stay on it.

PR #588, opened 2025-04-21
Add Cassandra integration

Balwancia:
Add Cassandra integration for ann-benchmark - single replica case


ann_benchmarks/algorithms/cassandra/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ann-benchmarks AS builder
+
+RUN apt-get update
+RUN apt-get install -y openjdk-11-jre-headless
+
+RUN apt-get install -y curl sudo
+ENV CASSANDRA_VERSION 4.1.0
```
swasik, line 7: For vector search we will need version 5.0

ann_benchmarks/algorithms/cassandra/Dockerfile
swasik, line 1: In you approach you try to base on ann-benchmarks Docker image but I think that it would be better to base on official Cassandra image similarly to what pgvector does. This way we have properly configured single node Cassandra working out of the box and adding run algorithms script is easy. Example Docker file:

```
FROM cassandra:5.0.4

# ----- OS packages --------------------------------------------------
RUN apt-get update && \
    DEBIAN_FRONTEND=noninteractive \
    apt-get install -y --no-install-recommends \
        python3 python3-pip && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*

# ----- Python dependencies ------------------------------------------
WORKDIR /home/app
COPY requirements.txt .
RUN pip install --no-cache-dir -r requirements.txt \
    && pip install --no-cache-dir cassandra-driver

# ----- Application code ---------------------------------------------
COPY run_algorithm.py .
COPY entrypoint.sh /usr/local/bin/
RUN chmod +x /usr/local/bin/entrypoint.sh

# ----- Ports --------------------------------------------------------
EXPOSE 7000 7001 7199 9042 9160

# ----- Entrypoint ---------------------------------------------------
ENTRYPOINT ["entrypoint.sh"]
```

And extra `entrypoint.sh` file that has to be created here:

```bash
#!/usr/bin/env bash
set -e                          # stop on the first error
trap 'kill -TERM $CASS_PID' EXIT

#
# 1.  Start Cassandra in the background
#     -R == “don’t run in the foreground” (contrary to -f)
#
docker-entrypoint.sh cassandra -R &
CASS_PID=$!

#
# 2.  Wait for CQL port 9042 to answer a simple query
#
echo "⏳ Waiting for Cassandra to come up …"
until cqlsh -u cassandra -p cassandra 127.0.0.1 9042 -e "DESCRIBE CLUSTER" >/dev/null 2>&1
do
  sleep 3
done
echo "✅ Cassandra is up!"

#
# 3.  Launch your application
#
python3 -u /home/app/run_algorithm.py "$@"

#
# 4.  Keep the container alive as long as Cassandra is running
#
wait $CASS_PID
```

swasik (changes requested):

ann_benchmarks/algorithms/cassandra/module.py
```diff
@@ -0,0 +1,64 @@
...
+        self.method_param = method_param
+        self.param_string = "-".join(k + "-" + str(v) for k, v in self.method_param.items()).lower()
+        self.index_name = f"os-{self.param_string}"
+
+        self.cluster = Cluster(self.config["host"])
+        self.conn = self.cluster.connect()
+        self._setup_keyspace()
+
+    def _setup_keyspace(self):
+        self.session.execute(f"""
```
swasik, line 20: I think you store the connection in `conn` member, not `session`.

ann_benchmarks/algorithms/cassandra/module.py
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        hnsw_distance_type = self.DISTANCE_MAPPING.get(self.metric, "EUCLIDEAN") 
+        self.conn.execute(f"""
+            CREATE INDEX IF NOT EXISTS {self.index_name}
+                ON {self.keyspace}.{self.table}(embedding) USING 'sai'
+                WITH OPTIONS = {{ 'similarity_function': '{hnsw_distance_type}' }};
+        """)
+        
+    def fit(self, X, batch_size=1000):
```
swasik, line 49: Batch size 1000 exceeds Cassandra limit. 100 works fine instead.

swasik (changes requested):

ann_benchmarks/algorithms/cassandra/module.py
```diff
@@ -0,0 +1,80 @@
...
+                WITH OPTIONS = {{ 'similarity_function': '{hnsw_distance_type}' }};
+        """)
+        
+    def fit(self, X, batch_size=100):
+        self.vector_dim = X.shape[1]
+        self._create_table()
+
+        insert_query = f"INSERT INTO {self.table_name} (id, embedding) VALUES (?, ?)"
+        prepared = self.conn.prepare(insert_query)
+        batch = BatchStatement()
```
swasik, line 56: Use `BatchStatement(batch_type=BatchType.UNLOGGED)` here - this will use unlogged batches which are faster (no need to write them to log which is not needed while using it just for benchmarking).

PR #598, opened 2025-06-18
Scylladb benchmark fixes

piodul:
Work in progress - usearch seems to work, but need to change it to use opensearch


piodul:
Sorry, submitted the PR to the wrong repo.

Issue #112, opened 2019-04-12
Add NSG


cmpute:
C++ code here: https://github.com/ZJULearning/nsg
I think providing wrapping with cython should be easy

erikbern:
sure – want to go ahead and give it a shot?

PR #611, opened 2025-09-19
Add Navigating Spreading-Out Graph (NSG)

twuebker:
NSG [1] is one of the state-of-the-art algorithms according to recent surveys [2] but still missing in the benchmark. 

Add NSG along with a very rudimentary configuration (based on what the authors said they used for gist and sift datasets) because NSG takes a bit longer to build compared to other indices. The NSG module uses Python bindings called pynsg that are a wrapper around the reference implementation [3].
Fix issue #112 

[1] https://arxiv.org/pdf/1707.00143
[2] https://dl.acm.org/doi/abs/10.1145/3709693
[3] https://github.com/ZJULearning/nsg

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -51,6 +51,7 @@ jobs:
           - nndescent
           - n2
           - nmslib
+          - nsg
           - onng_ngt
           - opensearchknn
           - panng_ngt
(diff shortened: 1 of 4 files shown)


twuebker:
@erikbern @maumueller Thanks for the great work on this benchmark. I'd really appreciate it if one of you could review and let me know if you want anything changed. Thanks!

PR #612, opened 2025-09-19
Fix DiskANN: Change to diskannpy

twuebker:
Currently, Diskann is disabled due to various issues, e.g. AVX512 dependence. The code used for the Diskann source and the Python bindings is from a branch of the diskann repository which has not been updated since 2021. Instead, diskannpy should be used, which are the official bindings and part of the main branch of Diskann. 

Refactor the diskann module to use the StaticMemoryIndex from the diskannpy library. Simplify Dockerfile and configuration, make the module minimal to reduce complexity, similar to the hnswlib module. Remove PQ since its primarily useful for indices that are stored on disk or take very long to build, and trades off recall [1].

[1] https://microsoft.github.io/DiskANN/docs/python/latest/diskannpy.html

### DIFF
--- a/ann_benchmarks/algorithms/diskann/Dockerfile
+++ b/ann_benchmarks/algorithms/diskann/Dockerfile
@@ -1,29 +1,9 @@
 FROM ann-benchmarks
 
 RUN apt-get update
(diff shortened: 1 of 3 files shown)


Michael-Jackson666:
This is a great change. Previously, files that called the vamanapy package couldn't run at all. Can they be merged?

PR #619, opened 2026-01-27
Add qingming-flat amd

uulong950:
### Implementation Details
- **Algorithm**: QingMing (AMD GPU Optimized)
- **Hardware**: Tested on RX 7900 XTX

### Verification
test passed locally.

qingming_official


uulong950:
are u ok?

PR #631, opened 2026-06-13
Add vecdb (PyPI: vecdbc)

navid72m:
vecdb is a from-scratch vector database in C11: HNSW graph with hand-written SIMD distance kernels (AVX-512 on x86, NEON on ARM). Published on PyPI as vecdbc; the Dockerfile installs it from there.

Adapter: ann_benchmarks/algorithms/vecdbc/ (module.py, config.yml, Dockerfile)
Handles euclidean directly and angular via L2-normalization
Single-threaded per the framework's convention
Tested locally: passes random-xs-20-euclidean (recall 1.0, exit 0 across all parameter groups) and runs cleanly on sift-128-euclidean
Added to CI test list for random-xs dataset

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -69,6 +69,7 @@ jobs:
           - tinyknn
           - vald
           - vearch
+          - vecdbc
           - vespa
           - voyager
           - vsag
--- a/ann_benchmarks/algorithms/vecdbc/Dockerfile
+++ b/ann_benchmarks/algorithms/vecdbc/Dockerfile
@@ -1,0 +1,14 @@
+FROM ann-benchmarks
+
+# vecdb is a C library compiled at install time; needs a C toolchain with
+# OpenMP. The sdist picks portable SIMD flags per arch (AVX2 on x86).
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        build-essential libgomp1 && \
+    rm -rf /var/lib/apt/lists/*
+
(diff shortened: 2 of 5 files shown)


navid72m:
<html><head></head><body><h2>Verification: recall/QPS on sift-128-euclidean</h2>
<p>Ran the adapter through the standard harness. Recall@10 vs the provided
ground truth, single-thread, M=16 / efConstruction=200, sweeping ef:</p>

ef | recall@10 | QPS
-- | -- | --
10 | 0.740 | 20,262
20 | 0.858 | 18,177
50 | 0.954 | 9,041
100 | 0.985 | 5,075
200 | 0.996 | 2,812
400 | 0.999 | 1,538


<p>(Numbers above measured on Apple Silicon; the Docker harness on x86 will use
the AVX2 build path. Posting these to show the adapter produces a correct,
monotonic precision/recall frontier — recall climbs smoothly with ef and
matches a reference HNSW implementation to ~3 decimals on this dataset.)</p>
<p>Build time on SIFT1M stays well under the "several hours" guideline
(~hundreds of seconds at these parameters). The config sweeps M ∈
{12,16,24,32,48} so the framework can pick the precision/performance frontier
rather than a single fixed point.</p>
<p>Also verified the angular path: vecdb uses squared-L2, so angular datasets are
handled by L2-normalizing both base and query vectors (cosine ranking ==
L2 ranking on unit vectors), giving correct recall against cosine ground
truth.</p>
<p>Happy to adjust parameters, the sweep, or anything else if it would make the
benchmark more useful.</p></body></html>

PR #634, opened 2026-08-04
Fix dimension inference in train_test_split when dimension=None

### LINKED ISSUE #610: Incorrect handling of dimension=None default parameter of train_test_split() in datasets.py
Location:
datasets.py
train_test_split()
line 206

Current: dimension = dimension if not None else X.shape[1]
Correct: dimension = dimension if dimension is not None else X.shape[1]

Since None is always False, the current behavior will leave dimension as None, rather than changing it to X.shape[1] when dimension=None as I assume was intended.

chakshu-dhannawat:
Closes #610.

train_test_split was meant to infer dimension from X.shape[1] when passed dimension=None, but the guard was written as `if not None` instead of `if dimension is not None`. Since `not None` is always True, the inferred value was never used and the log line always printed `*None`.

The fix changes the guard to `dimension is not None`, so callers can omit the argument and still see the correct dimensionality in the output.

Also added a regression test in `test/datasets_test.py` to verify both the default inference path and an explicit override.


chakshu-dhannawat:
Friendly ping for a review when you get a chance. Thanks.

